PR #12, opened 2015-02-03
build geoserver-2.6.2

danfruehauf:


danfruehauf:
Tested in vagrant, looks fine.

PR #13, opened 2015-02-20
new extension build process

julian1:


build.sh
```diff
@@ -0,0 +1,3 @@
+set -x
```
danfruehauf: You probably want the first line to be `#!/bin/bash`, then if you want `-x`, change it to `#!/bin/bash -x`

pom.xml
```diff
@@ -1,90 +1,56 @@
...
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
     <modelVersion>4.0.0</modelVersion>
     <groupId>au.org.emii</groupId>
     <artifactId>geoserver</artifactId>
-    <packaging>war</packaging>
-    <version>2.6.2</version>
+    <packaging>pom</packaging>
+    <version>1.0.0</version>
```
danfruehauf: You know what I think about the choice of version here.

I personally think it should reflect somehow the version of Geoserver that's encapsulated in the build.

build.sh
julian1: fixed!

build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -x
+git submodule update --init
+mvn clean install -U -Dmaven.test.skip=true
```
jonescc: This builds and installs to the local repository - did you want to do that?

pom.xml
```diff
@@ -1,90 +1,56 @@
...
+        </dependencies>
+    </dependencyManagement>
+
+     <!-- gs-xslt plugin -->
+    <repositories>
         <repository>
-            <id>${internal.repository.id}</id>
-            <url>${internal.repository.url}</url>
+            <id>opengeo</id>
+            <url>http://repo.boundlessgeo.com/main</url>
```
jonescc: So we've decided not to bother with Archiva for caching of artifacts?

danfruehauf:
@jonescc Please provide more code related feedback, I'll do some end to end testing of that with no_db, see how we go.

danfruehauf:
Alrighty, fine to be merged by me. @jonescc Any more comments?

jonescc:
No, apart from comments made, looks pretty good to me.  Nice work @julian1

danfruehauf:
So I'll just let @julian1 address the comments you've made.

build.sh
julian1: I almost removed this script file, as it's not used for Jenkins which is the correct place to build. So at the moment it's serves as a documented quickstart, to enable another dev picking this up to know what the required git submodule incantations are.

pom.xml
julian1: I would like to see this for maven projects. But feel this is probably best addresssed in a retro.

build.sh
julian1: @jonescc On seconds thoughts I might put this into the README

julian1: @jonescc updated.

danfruehauf:
I think it's all addressed. Can I merge?

PR #14, opened 2015-05-12
upgrade to geoserver 2.7.0

danfruehauf:
That works. Tested in dev box.

Committed also the change for the geoserver plugin in its repository. Anyway, GTG.


pmbohm:
I'll trust you

PR #15, opened 2015-05-28
build with ncdfgenerator

danfruehauf:
@julian1 I didn't test if it works. However it does include all the jars.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
 	path = src/extension/geoserver-layer-filter-extension
 	url = git@github.com:aodn/geoserver-layer-filter-extension
 	branch = master
+[submodule "src/extension/ncdfgenerator"]
+	path = src/extension/ncdfgenerator
+	url = git@github.com:aodn/ncdfgenerator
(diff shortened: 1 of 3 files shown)


danfruehauf:
Corresponding submodule branch:
https://github.com/aodn/ncdfgenerator/tree/geoserver2_fix_pom (forked from geoserver2)

danfruehauf:
OK, tested that it works. WPS is enabled and the service is published in `GetCapabilities`.

julian1:
I'm reviewing.

julian1:
@danfruehauf This successfully creates the WPS service. But the service fails when called because the ncdfgenerator dependencies such as the ucar netcdf jar are not being pulled in.

julian1:
hmm hang on, let me try again.

danfruehauf:
@julian1 didn't update here - but it does work.

Another thing we need to do, when running geoserver we need to specify `-Djava.awt.headless=true` otherwise I get:
```
java.lang.NoClassDefFoundError: Could not initialize class sun.awt.X11GraphicsEnvironment
```

PR #17, opened 2015-06-05
gogoduck wps support

danfruehauf:
I mean business.

By default this will be disabled as without config at `wps/gogoduck.xml` gogoduck will reject any incoming job.


jkburges:
This is dead now, right?

danfruehauf:
Dead as.

PR #18, opened 2015-06-18
Imos 2.7.0

julian1:


jonescc:
Looking at this as well

README.md
```diff
@@ -1,27 +1,32 @@
...
+* Netcdf output extension
 
-To build:
-
+### To build:
 ```
 git submodule update --init
-mvn clean install -U -Dmaven.test.skip=true 
+mvn clean
+mvn -P wps,xslt install -DskipTests
```
jonescc: No real reason to use install here - package will build all the artifacts.  We don't use local repositories for artifacts we build.

I'd remove -Dskiptests - we don't want to encourage building without testing.  Thats something the developer can choose to do if they want to speed up building while developing if they know what they are doing.

julian1:
fixed

jkburges:
Don't think this is relevant anymore.

PR #19, opened 2015-06-25
Update geoserver

julian1:


julian1:
hopefully the last one, and we can actually do some proper testing!

emii/pom.xml
```diff
@@ -90,8 +100,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.1</version>
                 <configuration>
-                    <source>1.5</source>
-                    <target>1.5</target>
+                     <source>1.5</source>
```
danfruehauf: What happened here?

emii/pom.xml
```diff
@@ -7,56 +7,66 @@
...
         <artifactId>geoserver</artifactId>
         <version>1.0.0</version>
-        <relativePath>../../pom.xml</relativePath>
     </parent>
 
     <dependencies>
+
+        <dependency>
+            <groupId>org.geoserver.web</groupId>
+            <artifactId>gs-web-app</artifactId>
```
danfruehauf: So why is this still here if we're taking geoserver as submodule?

julian1: It's either for the overlay manager which is still used, or for the custom xslt code.

emii/pom.xml
julian1: Maybe Craig's Eclipse?

emii/pom.xml
danfruehauf: Overlay manager?

danfruehauf:
I'm merging this, but I really want to be able to get back to source `gs-web-app` without the need of buliding geoserver. I hope we agree on that.

emii/pom.xml
julian1: yes.

julian1: We should get rid of it. But I don't want to play with it more right now.

danfruehauf: You assume I know what's the overlay manager. I don't.

What is it?

julian1: It munges the jars, to insert our custom class files and context for the csv with metadata header.

PR #21, opened 2015-10-13
Import gogoduck java implementation

danfruehauf:
* GGD worker ported to java (as maven project)
 * Works like the previous GGD
 * Can be invoked from command line
 * Produces a jar which can become a WPS service
 * Has some test coverage
 * Tested with the following
   * ACORN
   * SRS
   * CARS
   * GSLA


src/extension/gogoduck/doc/wps-acorn.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<wps:Execute service="WPS" version="1.0.0"
+  xmlns:wps="http://www.opengis.net/wps/1.0.0" xmlns:ows="http://www.opengis.net/ows/1.1"
+  xmlns:ogc="http://www.opengis.net/ogc" xmlns:xlink="http://www.w3.org/1999/xlink"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.opengis.net/wps/1.0.0 http://schemas.opengis.net/wps/1.0.0/wpsExecute_request.xsd">
+
+
```
pmbohm: spaces

danfruehauf: Will fix.

src/extension/gogoduck/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+ * nco tools, a recent version! (>= 4.3.4)
+ * java/maven
+
+## Usage:
+
+To build:
+```
+$ mvn clean install
+```
```
jkburges: I get [build errors](https://gist.github.com/jkburges/76171129d9e6101bccf9) when doing this.

src/extension/gogoduck/doc/wps-acorn.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
jkburges: In the three given example requests, the only difference appears to be the layer name and the TIME range.  Is it *really* necessary then to have the three examples, or would one be enough?  
Otherwise, it's more doc to keep up to date, more information for the reader to process.

danfruehauf: Fair enough, just thought it might come handy. I understand what you're saying. How should I arrange it then?

src/extension/gogoduck/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+    <artifactId>gogoduck</artifactId>
+    <version>2.0</version>
+
+    <parent>
+        <groupId>org.geoserver</groupId>
+        <artifactId>extension</artifactId>
+        <version>2.7.0</version>
+    </parent>
+
+    <dependencies>
```
jkburges: I think we're going to end up with a lot of duplication in the POM between the root POM and all of the `extension/*/pom.xml` files.  We should have a look at refactoring that once everything has been brought in to this repo.

src/extension/gogoduck/README.md
danfruehauf: Umn. I didn't quite understand which dependency it couldn't pull...

jonescc: Looks like they moved some dependencies to a geotools repository - I had to make changes to the geoserver-build project to pull them in now

danfruehauf: @jonescc Thanks for the pointer, I'll see with @jkburges how we can solve it.

jonescc: https://github.com/aodn/geoserver-build/pull/25/files#diff-600376dffeb79835ede4a0b285078036R54

src/extension/gogoduck/doc/wps-acorn.xml
jkburges: Well, I reckon just keep one.   Should be easy enough given one working example for people to adapt to whatever they need?

danfruehauf: I basically wanted to couple some sane parameters together, something that's easy to iterate on. So for ACORN if you use those parameters, you can a pretty quick response.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+        this.geoserver = geoserver;
+        this.profile = profile;
+        this.subset = subset;
+        this.outputFile = new File(outputFile).toPath();
+        this.limit = limit;
+        this.baseTmpDir = new File(System.getProperty("java.io.tmpdir")).toPath();
+        this.userLog = new UserLog();
+    }
+
+    public void setTmpDir(String tmpDir) {
```
jkburges: Any reason to have the settable?  Or can we just make them part of the constructor, and immutable?

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+    static {
+        replacePrefixes.put("/mnt/imos-t3/", "https://data.aodn.org.au/");
+        replacePrefixes.put("/mnt/opendap/2/SRS/", "https://thredds.aodn.org.au/thredds/fileServer/srs/");
+    }
+
+    public static final String ncksPath = "/usr/bin/ncks";
+    public static final String ncrcatPath = "/usr/bin/ncrcat";
+    public static final String ncdpqPath = "/usr/bin/ncpdq";
+
+    private final String geoserver;
```
jkburges: `geoserverUrl`?

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+        GoGoDuckModule module = getProfileModule(profile, geoserver, subset, userLog);
+
+        Path tmpDir = null;
+
+        try {
+            tmpDir = Files.createTempDirectory(baseTmpDir, "gogoduck");
+
+            URIList URIList = module.getUriList();
+
+            enforceFileLimit(URIList, limit);
```
jkburges: This is obviously the core high level algorithm, and I think it gives some clues as to how I *think* this (IMO) big class could be refactored.  I know we have differences of opinion in this area, but *IMO* adding some level of abstraction makes it easier for maintainers (e.g. you usually have less code to understand when making a change, as a more specific class is smaller).

I would guess at least specific classes for:

* downloading files (including enforcement of file limit)
* subsetting (+ post process)
* aggregating (+ update metadata)

would make sense.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+            if (isCancelled()) {
+                logger.warn("GoGoDuck was cancelled during download.");
+                return;
+            }
+
+            File srcFile = new File(uriList.get(0).toString());
+            String basename = new File(uri.toString()).getName();
+            Path dst = new File(tmpDir + File.separator + basename).toPath();
+
+            if(srcFile.exists() && !srcFile.isDirectory()) {
```
jkburges: What's the deal with this if/else? Can it be re-factored as such:

```java
if (srcFileExists()) {
  createSymbolicLink();
}
else {
  downloadFile();
}
```

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+
+            String extension = FilenameUtils.getExtension(dst.getFileName().toString());
+
+            if (extension.equals("gz")) {
+                gunzip(dst.toFile());
+            }
+        }
+    }
+
+    private static URL fileURItoURL(URI uri) {
```
jkburges: Maybe we can have a class which takes care of this ugliness, with all the relevant code there for people to see e.g.:

```java
public class UrlMangler {
  private static final Map<String, String> replacePrefixes ...

  public static URL mangle(URL url) {
    return magic();
  }
}
```

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+                }
+            }
+            return new URL(uriStr);
+        }
+        catch (MalformedURLException e) {
+            throw new GoGoDuckException(e.getMessage());
+        }
+    }
+
+    private static void gunzip(File file) {
```
jkburges: Maybe this method should be called `gunzipInPlace()`?

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+import org.apache.commons.io.FilenameUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.opengis.util.ProgressListener;
+
+import java.io.IOException;
+import java.util.*;
+import java.util.zip.GZIPInputStream;
+
+public class GoGoDuck {
```
jkburges: I wonder if `Worker` would be a better name for this class?  `GoGoDuck` is a bit vague, given that it sits in a "project" of the same name...

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+        catch (Exception e) {
+            throw new GoGoDuckException(String.format("Could not apply subset to file '%s': '%s'", file.getPath(), e.getMessage()));
+        }
+    }
+
+    private static void applySubsetSingleFileNative(File file, GoGoDuckModule module) {
+        // TODO implement!
+    }
+
+    private static class ncksRunnable implements Runnable {
```
jkburges: As mentioned above, I definitely think that this code to do with doing the actual subsetting could live in its own class (or classes, possibly one for the ncks specific stuff too).

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+    }
+
+    private void applySubsetMultiThread(Path tmpDir, GoGoDuckModule module, int threadCount) throws GoGoDuckException {
+        userLog.log(String.format("Applying subset '%s'", module.getSubsetParameters().toString()));
+        logger.info(String.format("Applying subset on directory '%s'", tmpDir));
+
+        File[] directoryListing = tmpDir.toFile().listFiles();
+        logger.info(String.format("Subset for operation is '%s'", module.getSubsetParameters()));
+
+        Deque<File> workQueue = new ArrayDeque<File>();
```
jkburges: `ArrayDeque` is not thread-safe, according to the doc.  [ConcurrentLinkedDeque](http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentLinkedDeque.html) *is* thread-safe, so maybe a more appropriate choice (if it still does what you want).

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
jkburges: In any case, I would capitialise to `NcksRunnable` to fit with java convention.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+            this.workQueue = workQueue;
+            this.module = module;
+            this.gogoduck = gogoduck;
+        }
+
+        public void run() {
+            try {
+                File file = null;
+                while ((file = workQueue.pop()) != null) {
+                    if (gogoduck.isCancelled()) {
```
jkburges: `isCancelled()` is not thread safe either, probably need to declare it as `synchronized`.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+
+        File[] directoryListing = tmpDir.toFile().listFiles();
+        logger.info(String.format("Subset for operation is '%s'", module.getSubsetParameters()));
+
+        Deque<File> workQueue = new ArrayDeque<File>();
+        for (File file : directoryListing) {
+            workQueue.push(file);
+        }
+
+        try {
```
jkburges: Have a look at [ExecutorService usage examples](http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html).  You may be able to simplify this code a bit, e.g.:

```java

...
// note - change NcksRunnable to operate on a single file, rather than pulling files off a queue, e.g. constructor such as:
ncksRunnable(String name, File file, GoGoDuckModule module, GoGoDuck gogoduck)
...

pool = Executors.newFixedThreadPool(threadCount);

  pool.execute(new NcksRunnable(..., file, ...);
}

pool.awaitTermination(...);
```

I *think* this code is equivalent - but hopefully you get the idea.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
danfruehauf: Declare `isCancelled` as thread safe?

It's a good question whether the `ProgressListener.isCancelled()` is actually thread safe, I'm not sure. I can declare it as synchronized to prevent ambiguity though.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+        command.add("-4");
+        command.add("-h");
+        command.add("-O");
+
+        File[] directoryListing = tmpDir.toFile().listFiles();
+        if (directoryListing.length == 1) {
+            // Special case where we have only 1 file
+            File file = directoryListing[0];
+            try {
+                if(outputFile.toFile().exists() && outputFile.toFile().isFile()) {
```
jkburges: `if (` (spacing)

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+    private static void cleanTmpDir(Path tmpDir) {
+        logger.debug(String.format("Removing temporary directory '%s'", tmpDir));
+        try {
+            FileUtils.deleteDirectory(tmpDir.toFile());
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+
+    private static GoGoDuckModule getProfileModule(String profile, String geoserver, String subset, UserLog userLog) {
```
jkburges: I think it'd make sense to move this to the `GoGoDuckModule` class itself.  It's typical to call "factory methods" such as this `newInstance`, so it would look like (up above):

```java
GoGoDuckModule module = GoGoDuckModule.newInstance(profile, geoserver, subset, userLog);
```

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -0,0 +1,511 @@
...
+        }
+        catch (InterruptedException e) {
+            logger.error(String.format("Interrupted: '%s'", e.getMessage()));
+            throw e;
+        }
+
+        return process.exitValue();
+    }
+
+    public class UserLog {
```
jkburges: Move to own file?

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
danfruehauf: Used by the main function, allows you to specify a different `tmpDir`. Other option is having another constructor, one with it and one without. I think that's less ugly.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
danfruehauf: I heat you. I'll have to leave that one for another PR though...

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -0,0 +1,176 @@
...
+
+public class GoGoDuckModule {
+    private static final Logger logger = LoggerFactory.getLogger(GoGoDuckModule.class);
+
+    protected String profile = null;
+    protected String geoserver = null;
+    protected SubsetParameters subset = null;
+    protected UserLog userLog = null;
+
+    public GoGoDuckModule() {}
```
jkburges: Any reason for using an `init()` method rather than just a constructor? 
Also, in classes which are going to be accessed from multiple threads, it's good practice to make them immutable, so in this case, change above to:

```java
 protected final String profile;
...
```

which also then forces them to be initialised in the constructor.  Maybe there's a reason for doing it the ways it's done here though.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
danfruehauf: `GoGoDuckWorker` maybe?

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/SubsetParameters.java
```diff
@@ -0,0 +1,47 @@
+package au.org.emii.gogoduck.worker;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+import au.org.emii.gogoduck.worker.SubsetParameters.SubsetParameter;
+
+public class SubsetParameters extends HashMap<String, SubsetParameter> {
```
jkburges: Does this class itself need to be thread-safe?  If I'm reading things correctly, I think not, as it's the one thread doing the writing, at the start, before the multiple threads simply *read* from here.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule_cars.java
```diff
@@ -0,0 +1,12 @@
+package au.org.emii.gogoduck.worker;
+
+public class GoGoDuckModule_cars extends GoGoDuckModule {
```
jkburges: Is this unconventional naming scheme to do with finding the correct module for a given layer?  I can live with it if so.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule_cars_australia_weekly.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+public class GoGoDuckModule_cars_australia_weekly extends GoGoDuckModule_cars {
+    @Override
+    public URIList getUriList() throws GoGoDuckException {
+        URIList uriList = new URIList();
+        try {
+            uriList.add(new URI("/mnt/imos-t3/climatology/CARS/2009/eMII-product/CARS2009_Australia_weekly.nc"));
```
jkburges: Does any of this need to be configurable?

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
danfruehauf: It's because of the way I use to instantiate those classes. So getting a ctor with no parameters and then calling `init()`. Usually for plugin design patterns is what will happen. With java I can probably do it also with ctor arguments, but the reflection code will be ugly :)

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/Main.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        if (cmd.hasOption("S")) {
+            ggd.score();
+        }
+        else {
+            ggd.run();
+        }
+    }
+}
```
jkburges: intellij? :smile:

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule_cars.java
danfruehauf: Yep. So that corresponds to the specialization you need for specific layers. So if you need specialization for `cars_2009`, you can just build:
```
public class GoGoDuckModule_cars_2009 extends GoGoDuckModule_cars
```

Then have specific behaviour for that. It's not very java-y, but I think it's not /that/ bad...

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule_cars_australia_weekly.java
danfruehauf: No... And it should be! I'll do it. Now that I'm an expert with dependency injection ;)

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/SubsetParameters.java
danfruehauf: `getNcksParameters` needs to be thread safe. However it's not modifying the map, just iterating on it. I think it should be OK. No such thing as `const` functions in java, right?

src/extension/gogoduck/src/test/java/au/org/emii/gogoduck/worker/GoGoDuckModuleTest.java
```diff
@@ -0,0 +1,72 @@
...
+        protected List<Attribute> getGlobalAttributesToUpdate(NetcdfFile nc) {
+            List<Attribute> newAttributeList = new ArrayList<Attribute>();
+            newAttributeList.add(new Attribute("test_attribute", "test_value"));
+            return newAttributeList;
+        }
+    }
+
+    @Test
+    public void testUpdateMetadata() throws Exception {
+        File sampleNetcdf = new File("src/test/resources/file.nc").getAbsoluteFile();
```
jkburges: Looking at this test has made me realise that many java APIs tend to deal with streams, rather than files, for added flexibility (and probably performance).  So here for example, you could avoid dealing with real files, and use say `ByteArrayOutputStream` - faster (important for unit tests), and cleaner, because you don't have to cleanup after yourself.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
jkburges: Oh yeah I see.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
jkburges: I think the `GoGoDuck` bit is redundant, since anyone using this class would already be working in that context.  It's ok if you like it though!

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
jkburges: Yeah sorry, I meant:

```java
 public synchronized boolean isCancelled() {
...
}
```

You need to make this method thread-safe, since it's checking if `progressListener` is null.  Then it doesn't really matter if `ProgressListener.isCancelled()` is thread-safe.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
jkburges: Ah yeah, it makes sense now. I thought there'd be a reason.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/SubsetParameters.java
jkburges: No but you might be able to make use of http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableMap(java.util.Map) or the like.

src/extension/gogoduck/doc/wps-acorn.xml
danfruehauf: What do you reckon for the doc then?

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
danfruehauf: This became thread safe (it's synchronized).

src/extension/gogoduck/doc/wps-acorn.xml
jkburges: I'm not going to lose sleep over this.  You're right, it might be useful to have a handful of examples that work out of the box. Leave as is for now.

jkburges:
Just tested the examples given in the README and they all worked. :astonished:

danfruehauf:
@jkburges NICE!!! :+1: 

I'm also quite :astonished:, but it's a good sign I believe. Hope it goes smooth also when it becomes WPS!!

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/Main.java
danfruehauf: Funny github doesn't show this line was modified. I added a new line, but probably because the line after was modified. Hmmmm.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/SubsetParameters.java
danfruehauf: So when iterating on the map, I use `unmodifiableMap(this)`, which returns a new copy of the map. Speaking of which, also a simple copy of the map or simply making the function `synchronized` will work.

Looking at that again, I'll just make the function synchronized. It's simpler and less likely to cause eyebrows to raise.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -0,0 +1,212 @@
...
+        subsetParametersNoTime.remove("TIME");
+        return subsetParametersNoTime;
+    }
+
+    public static GoGoDuckModule newInstance(String profile, String geoserver, String subset, UserLog userLog) {
+        String thisPackage = GoGoDuckModule.class.getPackage().getName();
+        String classToInstantiate = String.format("GoGoDuckModule_%s", profile);
+
+        GoGoDuckModule module = null;
+        while (null == module && "" != classToInstantiate) {
```
jkburges: `"" != classToInstantiate ` will always evaluate to `true` in java.

Could be re-written as:

```java
while (null == module && !classToInstantiate.isEmpty())
```

src/extension/gogoduck/src/test/java/au/org/emii/gogoduck/worker/GoGoDuckModuleTest.java
danfruehauf: So as discussed, we'll leave it for later. This is a welcomed change and this is probably the correct way of doing that. The cruxes are that there are a few "passes" (multiple updating operations) on the file.

Another consideration is the result file. Can we have a 30GB `ByteOutputStream`? Good question, I'm not sure how java will handle it...

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
dnahodil: There is a method which gets a Constructor for a Class based on the arguments for the constructor. Might help you do this if you wanted to give it a go.
https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getConstructor-java.lang.Class...-

src/extension/gogoduck/src/test/java/au/org/emii/gogoduck/worker/GoGoDuckModuleTest.java
jkburges: > The cruxes are that there are a few "passes"

TBH, I'm not sure about how this would work with streams.

> Can we have a 30GB ByteOutputStream

I would say not, but in production code, you would probably want to use a `FileOutputStream` (if expecting to deal with such large amounts of data).  In tests, use a `ByteOutputStream` (since minimal data, faster, cleaner).

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
danfruehauf: Yeah, this is what I said. Java allows you to do it, but the code will be ugly. I tried it when writing the GN link monitor, then looked at it again and was like "WTF IS THAT?!", quickly factored it into a simple `init()` function. I think by now I completely removed that code, but tell you what, it did not look good :)

src/extension/gogoduck/src/test/java/au/org/emii/gogoduck/worker/GoGoDuckModuleTest.java
danfruehauf: > I would say not, but in production code, you would probably want to use a FileOutputStream (if expecting to deal with such large amounts of data). In tests, use a ByteOutputStream (since minimal data, faster, cleaner).

:+1: Agreed.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
danfruehauf: Something along the lines of:
```
Constructor cons = obj.getClass().getConstructor(new Class[] {org.jdom.Element.class, org.java.lang.String.class});
```

You get the idea... It's just an example.

dnahodil: I don't think you should need an array of Classes. Something like: (if you moved the init() args to the constructor)
`Constructor cons = obj.getClass().getConstructor(String.class, String.class, String.class, UserLog.class);`

But it's up to you.

danfruehauf: I think I'll leave it like that. Then I need to modify constructors in all inherited classes etc to call the parent with those parameters.

danfruehauf:
@jkburges @dnahodil Give this thing another go, I addressed the most painful parts except for one - that is the configurable URL mangler, which I'm doing now.

And BTW, this code will need more kneading. I tried to keep the logic as close as possible to what was in the original Bash implementation, however it doesn't mean it has to stay like that.

danfruehauf:
1a58be1 externalizes the config of the URL mangling. I haven't tested it in WPS mode (when running in geoserver). I suspect it should be OK. I can give it a test once the geoserver build is fixed. I know @jonescc is working on that.

jkburges:
Just the couple of small questions/comments on those last couple of commits.  Just tested again, all good.

src/extension/gogoduck/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule_cars_australia_weekly.java
danfruehauf: Oh wait, this needs to be configurable!!

jkburges:
Don't worry about the travis failure, this branch doesn't have the appropriate travis config yet.

danfruehauf:
@jkburges I can rebase!

danfruehauf:
Rebased against master.

@jkburges Please have a look at  3b9fa96 - this makes the CARS URIs sort of configurable.

PR #24, opened 2015-10-14
Remove GenerationIT tests

danfruehauf:
If one wants to run ncdfgenerator it can be done from the command line
specifying all needed credentials. This just will just fail if those
environment variables are unset.


jonescc:
These look like very useful integration tests.   Would it be better if we just didn't run them by default?  At least then they could be run locally where a postgres instance is installed or perhaps in an integration test environment at some point in the future.

danfruehauf:
@jonescc I know they are useful, but at this stage, it's like... Meh. The other thing we could do is disable the tests here.

AFAIK @julian1 created them so he can test the generation of files. Eventually those are not quite tests because they don't compare equality etc. And those things can be achieved using `mvn exec` with the main class.

jonescc:
It looks like a mix to me of things that compare equality and things that don't (but should)

Its simple enough to only run them when you want to:

See http://www.petrikainulainen.net/programming/maven/integration-testing-with-maven/

jonescc:
Or to just say - it doesn't matter if they fail

danfruehauf:
Not sure I'll have time to look at that if we want it to be this way, sorry.

PR #25, opened 2015-10-14
Upgrade to Geoserver 2.8.0 and fix missing dependencies

jonescc:


src/extension/geoserver-layer-filter-extension/pom.xml
```diff
@@ -14,7 +14,7 @@
     <parent>
         <groupId>org.geoserver</groupId>
         <artifactId>extension</artifactId>
-        <version>2.7.0</version>
+        <version>2.8.0</version>
```
jonescc: Would prefer to use our extensions module - will look at whether we can do this easily separately

danfruehauf:
I'm looking into that.

Will also rebase the other 2 PRs to match this.

danfruehauf:
@jonescc Works for me. Now we need to make it work for anyone...

PR #26, opened 2015-10-14
Fix dependency issues/version conflicts/startup problems

jonescc:
Use our own extension module as parent for extensions to simplify dependency management.


jonescc:
PS.  Also means that to upgrade we only need to change the geoserver version number in the  main pom.  All dependencies used in extensions will be for that version unless we specify otherwise.

jkburges:
@jonescc, I'm still having trouble building a war (although getting further than previously):

```
$ mvn clean install -DskipTests=true
```

Result: https://gist.github.com/f448838058c5b4d62c64

jkburges:
Oh hang on, I didn't skip tests properly.

danfruehauf:
Due to ncdfgenerator tests. We need to disable just them.
```
mvn clean install -Dmaven.test.skip=true
```

jkburges:
... but still failing:

https://gist.github.com/jkburges/6447d922f1413301f1a8

jonescc:
@jkburges and @pmbohm - have fixed build issues and startup problems identified above and demonstrated.

jkburges:
Good work!

PR #31, opened 2015-10-15
Import WPS notifier

jkburges:


src/extension/notifier/src/main/java/au/org/emii/wps/HttpNotifier.java
```diff
@@ -0,0 +1,39 @@
+package au.org.emii.wps;
+
+import java.net.URL;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.geotools.data.ows.HTTPClient;
+import org.apache.http.client.utils.URIBuilder;
+import java.io.IOException;
+import java.net.URISyntaxException;
```
dnahodil: This import statement is unused

src/extension/notifier/src/main/java/au/org/emii/wps/HttpNotifier.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class HttpNotifier {
+    private static final Logger logger = LoggerFactory.getLogger(NotifierProcess.class);
```
dnahodil: This logger is unused, it could be removed (if you do remove it there's a few import statements that can go, too)

src/extension/notifier/src/main/java/au/org/emii/wps/HttpNotifier.java
```diff
@@ -0,0 +1,39 @@
...
+        String uuid,
+        String notificationParams
+    ) throws IOException {
+        try {
+            URIBuilder builder = new URIBuilder(notificationUrl.toURI());
+            builder.setParameter("server", wpsServerUrl.toExternalForm());
+            builder.setParameter("uuid", uuid);
+
+            String callbackUrlAsString = builder.build().toURL().toExternalForm();
+            callbackUrlAsString += "&" + notificationParams;
```
dnahodil: It's a shame we need to use String concatenation here and can't use the URIBuilder class for this as well. We could use URIBuilder if `notificationParams` was a List of NameValuePairs but I assume that won't fit well with the calling code.

src/extension/notifier/src/test/java/au/org/emii/wps/HttpNotifierTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.net.URL;
+import java.net.MalformedURLException;
+import org.geotools.data.ows.SimpleHttpClient;
+import org.geotools.data.ows.HTTPClient;
+
+import static org.junit.Assert.*;
```
dnahodil: The Before, MalformedURLException, and Assert.* imports are unused.

src/extension/notifier/src/test/java/au/org/emii/wps/NotifierProcessTest.java
```diff
@@ -0,0 +1,53 @@
+package au.org.emii.wps;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.geoserver.wps.resource.WPSResourceManager;
+import org.geoserver.wps.process.RawData;
+
+import java.net.URL;
+import java.net.MalformedURLException;
```
dnahodil: Unused

src/extension/notifier/src/main/java/au/org/emii/wps/HttpNotifier.java
```diff
@@ -0,0 +1,39 @@
...
+            httpClient.get(new URL(callbackUrlAsString));
```
dnahodil: Should this maybe have a retry system in case of temporary network problems or something?

As it stands, I think, if something isn't quite right on the first attempt then the notification will never happen.

dnahodil:
This looks good, pretty straightforward. :+1:

Most of my comments are simple, probably only the last one needs some thought.

src/extension/notifier/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.5.1</version>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.11</version>
```
danfruehauf: Remove the version here.

src/extension/notifier/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>1.8.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>1.6.1</version>
```
danfruehauf: And here too, it comes from upstream.

src/extension/notifier/pom.xml
dnahodil: Just removing the version broke the build. Is there something it needs to be replaced with?

danfruehauf: So just leave it, we'll sort it out later.

It'll need to be rebased, so when it's done, we'll deal with it.

It also has `<scope>test</scope>` so it should be OK.

src/extension/notifier/src/main/java/au/org/emii/wps/NotifierProcess.java
```diff
@@ -0,0 +1,62 @@
...
+            return notifiableData;
+        }
+        catch (IOException e) {
+            logger.error("Error sending notification", e);
+            throw new ProcessException(e);
+        }
+    }
+
+    URL getWpsUrl() throws MalformedURLException {
+        // TODO is there a nicer way of getting BaseUrl?
```
danfruehauf: Haha! Stolen from GGD! I hope we fix it sooner than later :)

src/extension/notifier/src/main/java/au/org/emii/wps/HttpNotifier.java
jkburges: I agree - I tried that approach, but couldn't find a way to pass in a map as a WPS input.  That's not to say it *isn't* possible though.

src/extension/notifier/src/main/java/au/org/emii/wps/HttpNotifier.java
jkburges: I agree with this also, but that's for version 2.0 IMO.  I'd rather see the happy path working end to end first.  But I can add it to my list of TODO which hopefully I'll get to this sprint.

src/extension/notifier/src/main/java/au/org/emii/wps/HttpNotifier.java
danfruehauf: It's a good question of whether this code should break down the passed parameters and then rebuilds them. If it does it, it'll probably reduce the risk of dangerous injections. However if it doesn't it's also OK I assume.

jkburges:
I'm just having a look at those couple of pom issues, and testing that this actually still works, after rebasing on to master.

src/extension/notifier/pom.xml
jkburges: This dependency is *not* specified in the parent pom(s) (which is why removing the version broke things).  But perhaps junit is used by enough component to be worthy of having its dependency in the top level pom?  The advantage is less code + every component using the same version of junit.  Disadvantage is more complicated root pom, doesn't apply to *all* components.

Thoughts @jonescc?

danfruehauf: See:
https://github.com/aodn/geoserver-build/blob/master/src/extension/gogoduck/pom.xml#L37

Because it's included in a higher level pom, you do not need to include the version, this is what @jonescc told me at least :)

But rebase against master before that.

jkburges: Removing the version for junit works for me.  Removing slf4j version breaks.

danfruehauf: Possibly slf4j-simple can go to a top level pom, then we will not need to specify it everywhere. Can I amend this PR?

jkburges: I'm not sure if that's the right thing to do.  @jonescc?  Also, I want to understand what's going on with maven. If junit version is being pulled in from upstream (somewhere?) then why is it necessary to specify the dependency here at all?

danfruehauf: It is necessary so this maven project can build & run. If you specify the version, it'll try to run with a specific version ( @jonescc  taught me that).

jkburges:
This PR can now be built as a WAR and dropped in to tomcat, where it works.

Please, can we sort the pom stuff out on a separate PR, I don't want to be held up by that.

danfruehauf:
I'll sort the pom-ing out after this is merged.

jkburges:
Thanks @danfruehauf.

src/extension/notifier/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.11</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>1.8.5</version>
```
danfruehauf: Needs `<scope>test</scope>` so it doesn't package it in the jar.

jkburges: Done.

Issue #32, opened 2015-10-15
Netcdf generator merge


jkburges:
It seems as though we've got two "working copies" of ncdfgenerator:

* https://github.com/aodn/geoserver/commits/imos-2.7.1.1/src/community/ncdfgenerator
  * deployed to systest and production
* https://github.com/aodn/DELETE_ME.ncdfgenerator
  * deployed to ???
  * merged in to this project
  * won't be available as a WPS service within the geoserver war, built and run from this project, due to some missing files (see https://github.com/aodn/geoserver/commits/imos-2.7.1.1/src/community/ncdfgenerator)
  * includes some bug fixes not included above

We need to merge the two forks above to get a working WPS/netcdfgenerator, including bug fixes, within geoserver.  The portal integration task depends on this.

danfruehauf:
Hopefully it'll fix everything:

https://github.com/aodn/geoserver-build/pull/33



________________________________

jkburges:
I'm happy for this to be closed now :grinning: Well done everyone.

PR #33, opened 2015-10-15
Fix mayhem

danfruehauf:
* Import Julian changes (includes WPS implementation)
   * Make CLI work after that (tested!)
 * Fix pom.xml indent
 * Remove redundant repositories from plugins (use repositories from main pom.xml)

Phew.


jonescc:
I'll review.  Travis not working for this repo?

danfruehauf:
Ah, should work. I'll have a quick look.

danfruehauf:
Just pushed another commit to fix the version of ncdfgenerator (modified to 1.0.0).

danfruehauf:
Travis is good now!

jonescc:
I get a test failing when building locally:

Failed tests:   testGetNcksParameters(au.org.emii.gogoduck.worker.SubsetParametersTest): expected:<PARAM[2,2,3]> but was:<PARAM[3,4,5]>

Have to leave this for now.  Someone else can pick this up tommorrow

danfruehauf:
Possibly changing `SubsetParameters` to inherit from `LinkedHashMap` will solve it, as the order will be kept. I'll need to setup java 8 to test it though.

danfruehauf:
Fixes #32

jkburges:
Can I suggest that instead of changing the code to make tests work (if ordering isn't actually needed), change the test to check that map a "contains all" elements from map b, and not more.

Caveat: replying from my phone, haven't looked at the code, but hopefully it makes sense!

danfruehauf:
So it returns a list of arguments, something along the lines of `-d something1 -d something2`.

It doesn't matter whether it's `-d something1 -d something2` or `-d something2 -d something1`, but I want the `-d` in the right places.

Hope it makes sense.

PR #37, opened 2015-10-16
WPS notifier retry functionality

dnahodil:
This adds retry functionality to the WPS notifier.

The only thing with this is that say you try to notify 3 times and they all fail, we only act on the last exception. So if the exceptions are different we don't really find out about that. Not a big deal though, IMO.

### DIFF
--- a/src/extension/notifier/src/main/java/au/org/emii/wps/NotifierProcess.java
+++ b/src/extension/notifier/src/main/java/au/org/emii/wps/NotifierProcess.java
@@ -22,10 +22,14 @@ public class NotifierProcess implements GeoServerProcess {
 
     private final WPSResourceManager resourceManager;
     private final HttpNotifier httpNotifier;
+    private final int maxNotificationAttempts;
+    private final int retryInterval;
 
-    public NotifierProcess(WPSResourceManager resourceManager, HttpNotifier httpNotifier) {
+    public NotifierProcess(WPSResourceManager resourceManager, HttpNotifier httpNotifier, int maxNotificationAttempts, int retryInterval) {
         this.resourceManager = resourceManager;
         this.httpNotifier = httpNotifier;
+        this.maxNotificationAttempts = maxNotificationAttempts;
+        this.retryInterval = retryInterval;
     }
 
     @DescribeResult(name="result", description="NetCDF file", meta={"mimeTypes=application/x-netcdf"})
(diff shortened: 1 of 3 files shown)


src/extension/notifier/src/main/resources/applicationContext.xml
```diff
@@ -4,6 +4,8 @@
     <bean id="notifierProcess" class="au.org.emii.wps.NotifierProcess">
         <constructor-arg index="0" ref="wpsResourceManager"/>
         <constructor-arg index="1" ref="httpNotifier"/>
+        <constructor-arg index="2" value="3" />
+        <constructor-arg index="3" value="300000" />
```
danfruehauf: May I suggest injecting that as a property? Then it'll have a name and be more self-describing. See also:
https://github.com/aodn/geoserver-build/blob/ncwms_wps_plugin/src/extension/ncwms/src/main/java/applicationContext.xml#L8

*The presenter is not a spring expert*

dnahodil: Names are nicer than argument indexes. I'll have a look.

danfruehauf: You'll just need to have a setter and getter for each. There also must be an annotation for that. If there *is* an annotation for that, I'll change GGD and ncwms accordingly.

src/extension/notifier/src/main/java/au/org/emii/wps/NotifierProcess.java
```diff
@@ -39,8 +43,31 @@ public RawData execute(
...
+                }
+                catch (IOException e) {
+                    logger.info("Attempt #" + attempt + " failed", e);
+                    lastException = e;
+
+                    try {
+                        Thread.sleep(retryInterval);
+                    }
+                    catch (InterruptedException ie) {
+                        logger.debug("Sleep interrupted", ie);
```
jkburges: Story of my life.

src/extension/notifier/src/main/java/au/org/emii/wps/NotifierProcess.java
```diff
@@ -39,8 +43,31 @@ public RawData execute(
     ) throws ProcessException {
 
         try {
-            httpNotifier.notify(callbackUrl, getWpsUrl(), getId(), callbackParams);
-            return notifiableData;
+            IOException lastException;
```
jkburges: I think I'd rather see this logic go in the `HttpNotifier` class, e.g. an overloaded method `notify()`, which also takes as parameters the `maxNotificationAttempts` and `retryInterval`.  This class then remains as just "plumbing" of the interface between geoserver/WPS and our custom code.

jkburges:
Why closing? Forget what I said before about using a library to do this, I think this is ok!

danfruehauf:
He wants to merge into master, that's why...

dnahodil:
Sorry to pull the rug out from under you @jkburges.
https://github.com/aodn/geoserver-build/pull/40

PR #38, opened 2015-10-18
Move slf4j to main pom.xml

danfruehauf:
Reused by a few modules


danfruehauf:
I'll need to rebase after:
https://github.com/aodn/geoserver-build/pull/31

So wait with the merge.

PR #40, opened 2015-10-19
WPS notifier retry

dnahodil:
@jkburges thought we might be able to use an Apache library to achieve the same result I'm happy to go down that path and will just leave it up to @jkburges.

Also, I'm thinking that these new parameters should not be passed into the constructor but maybe be passed into the execute method. But this will only be nice if you can configure a default so that they're not required all the time (possibly can be done with the DescribeParameter annotation).

### DIFF
--- a/src/extension/notifier/src/main/java/au/org/emii/wps/NotifierProcess.java
+++ b/src/extension/notifier/src/main/java/au/org/emii/wps/NotifierProcess.java
@@ -22,10 +22,14 @@ public class NotifierProcess implements GeoServerProcess {
 
     private final WPSResourceManager resourceManager;
     private final HttpNotifier httpNotifier;
+    private final int maxNotificationAttempts;
+    private final int retryInterval;
 
-    public NotifierProcess(WPSResourceManager resourceManager, HttpNotifier httpNotifier) {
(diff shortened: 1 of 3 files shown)


src/extension/notifier/src/main/java/au/org/emii/wps/NotifierProcess.java
```diff
@@ -39,8 +43,31 @@ public RawData execute(
     ) throws ProcessException {
 
         try {
-            httpNotifier.notify(callbackUrl, getWpsUrl(), getId(), callbackParams);
-            return notifiableData;
+            IOException lastException;
```
jkburges: See https://github.com/aodn/geoserver-build/pull/37#discussion_r42329152

danfruehauf:
Perhaps careful with allowing WPS to describe those parameters, they must have upper limits. The reasons are obvious I believe.

src/extension/notifier/src/main/java/au/org/emii/wps/NotifierProcess.java
dnahodil: I'll have a look and see how that might look.

dnahodil:
Yeah an upper limit on retry count and and an upper and lower limit on retry interval would be required.

dnahodil:
New PR: https://github.com/aodn/geoserver-build/pull/44

PR #41, opened 2015-10-19
More pom cleanups

danfruehauf:
Minor modifications to get things a bit cleaner.

Notice I've also commented out the ncdfgenerator `GenerationIT` tests, this is because we should be able to run `mvn test install`.


danfruehauf:
Referenced by #42, I don't think we quite need a tag...

PR #43, opened 2015-10-19
Fix so that the filters extension is available

jkburges:
Not sure how it was working before with the file in the non-standard place, but anyway, fixed now.


danfruehauf:
:+1:

PR #47, opened 2015-10-19
Allow chained WPS processes to execute even after failure of NetCDF g…

jkburges:
…eneration

### DIFF
--- a/src/extension/ncdfgenerator/src/main/java/au/org/emii/wps/NetcdfOutputProcess.java
+++ b/src/extension/ncdfgenerator/src/main/java/au/org/emii/wps/NetcdfOutputProcess.java
@@ -4,7 +4,9 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.ByteArrayInputStream;
 import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.sql.Connection;
 import java.sql.SQLException;
 
@@ -24,6 +26,7 @@ import org.geotools.process.factory.DescribeResult;
 
 import org.geoserver.wps.gs.GeoServerProcess;
 import org.geoserver.wps.process.FileRawData;
+import org.geoserver.wps.process.RawData;
 import org.geoserver.wps.resource.WPSResourceManager;
 import org.geoserver.wps.ProcessDismissedException;
 import org.geoserver.catalog.Catalog;
@@ -67,7 +70,7 @@ public class NetcdfOutputProcess implements GeoServerProcess {
 
     @DescribeResult(name="result", description="Zipped netcdf files", meta={"mimeTypes=application/zip"})
 
-    public FileRawData execute(
+    public RawData execute(
         @DescribeParameter(name="typeName", description="Collection to download")
         String typeName,
         @DescribeParameter(name="cqlFilter", description="CQL Filter to apply")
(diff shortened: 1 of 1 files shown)


danfruehauf:
Should we do the same also for GGD? Not sure how to share this kind of code though...

jkburges:
Actually, yes @danfruehauf - perhaps I can do better than this, by introducing a wrapper WPS job which will execute a sub job, and on failure, return an error file.  Then it'll be up to the client to construct a query to use it.

danfruehauf:
@jkburges Didn't want to complicate things, it's just that those things will need to be done for both, so we can prevent duplicated code. If needed you can also introduce another extension which will have common WPS helper code for our extensions (and used by both GGD and JODAAC).

danfruehauf:
Delete that branch?

jkburges:
Umm can do.  Was going to use it soon so didn't bother.

jkburges:
Note to self: make sure that the Notifier process status "inherits" the upstream status, so that the portal job report page shows the status of the WPS process we're interested in, rather than the status of the Notifier.

jkburges:
I might possibly re-open this (or something using the same approach) by the end of today if I don't get anywhere with coming up with a better solution.  At least we will have something which can be released and tested.

jkburges:
Ok, here's what I'm going to do.  I will create a new Process type, let's call it `Fallible`.  This process will wrap one other WPS process, forwarding any inputs and likewise outputs - but in the case of the wrapped process failing, it will itself not fail, but just set the output to the error message/exception from the wrapped process.

The client will maintain flexibility over when to use `Fallible`, and what the wrapped process is (e.g. `NcdfGenerator` or `GoGoDuck`).

An example request might look like https://gist.github.com/jkburges/e3be577c52a11793166a.

I think this is the best approach for now - I think it's fairly easy to achieve this without having to dig in to geoserver/WPS code and/or make too many alterations to what we already have in our own extensions.

Thoughts @jonescc @danfruehauf ?

danfruehauf:
@jkburges That sounds good to me. Just the name, umn, uh. `Fallible`?

Well, yep, OK, `Fallible` :smile:

jkburges:
> Just the name, umn, uh. Fallible

http://www.thefreedictionary.com/fallible - got something better?! :confused:

jonescc:
Sounds good to me too.

jonescc:
Is this for errors the process can't catch itself?

jkburges:
> Is this for errors the process can't catch itself?

Not really.  If you look at this PR, I was originally going to change the `NcdfOutputProcess` to catch all exceptions and return an error in a file as the output.  But then, as @danfruehauf pointed out, I'd have to make the same change to the `GoGoDuckProcess` - so I thought it'd be better (less duplication, more flexible) to do what's described at  https://github.com/aodn/geoserver-build/pull/47#issuecomment-151680566

jonescc:
Could that duplication be sorted by inheritance or composition now your combining the modules?

jonescc:
OK, I see you've discussed this before - I think the api has been complicated a little and I like clean apis but its not a big deal.  Sorry to go over old ground.

jkburges:
> Could that duplication be sorted by inheritance or composition now your combining the modules?

Yes, it could.  But I think the solution I'm proposing is essentially composition but it's the client that's doing the composition, rather than hard-coded, if that makes sense.

I'm not sure what you mean by the API being complicated?  Do you mean the the request the client sends will be more complicated?

jonescc:
OK.  No problem.  And yes I was talking about the request"

Issue #45, opened 2015-10-19
WPS netcdfgenerator limits


jkburges:
Currently, the `NetcdfOutputProcess` requires a cql filter to be provided in the request.  Conversely, the portal enforces no such behaviour, meaning that it's possible for a user to request a "Download Later" download which will error.

There are two options:

1. Make the cql filter optional on the backend
1. Enforce that a collection is subsetted in the portal before requesting to "Download Later"

Note: even if no filter is provided, the WPS time execution limit will protect against too large requests.

jkburges:
@pblain thoughts?  I think this needs to be resolved before WPS is released in portal, as it's pretty easy for users to submit something that will fail as it stands currently.

danfruehauf:
Regarding the row limit, let the WPS time execution limit take care of that - and it already does.

jkburges:
Oh ok, I forgot about that.  So, people agree that the best option is (1) from above?

danfruehauf:
I give option (1) a :+1: - more generic.

pblain:
@jkburges, yes option 1 is my preference. Seems simpler and I don't see a downside.

PR #48, opened 2015-10-19
CQL filter is now optional

jkburges:
Fixes https://github.com/aodn/geoserver-build/issues/45
It was previously quite easy for the portal to construct invalid WPS netcdf requests.

### DIFF
--- a/src/extension/ncdfgenerator/src/main/java/au/org/emii/ncdfgenerator/Main.java
+++ b/src/extension/ncdfgenerator/src/main/java/au/org/emii/ncdfgenerator/Main.java
@@ -25,7 +25,7 @@ public class Main {
         options.addOption("d", "db", true, "Database connection string.");
         options.addOption("D", "driver", true, "Database driver class.");
         options.addOption("s", "schema", true, "Database schema to use.");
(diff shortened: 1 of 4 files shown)


danfruehauf:
So looks good, perhaps also modify the command line interface to allow an empty filter:
https://github.com/aodn/geoserver-build/blob/master/src/extension/ncdfgenerator/src/main/java/au/org/emii/ncdfgenerator/Main.java

jkburges:
> modify the command line interface to allow an empty filter

Good pickup - done.

A couple of notes:

* there's a bit of duplication between https://github.com/aodn/geoserver-build/blob/master/src/extension/ncdfgenerator/src/main/java/au/org/emii/ncdfgenerator/Main.java and https://github.com/aodn/geoserver-build/blob/master/src/extension/ncdfgenerator/src/main/java/au/org/emii/wps/NetcdfOutputProcess.java - might be nice to factor that out at some stage
* noticed https://github.com/aodn/geoserver-build/issues/49 while testing this PR - but I *think* it's unrelated to this actual PR, except for the fact that this PR exposes the problem.

danfruehauf:
@jkburges Now that we're on the correct development branch, we can easily fix them ;)

PR #55, opened 2015-10-21
Parse styles from get capabilities XML

danfruehauf:


danfruehauf:
@pmbohm I was thinking more along these lines. This is significantly less code.

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/Ncwms.java
```diff
@@ -114,45 +113,39 @@ private String encodeMapForRequest(Map<String, String[]> params) {
...
-            add("linevec");
-            add("vector");
-            add("boxfill");
-        }};
-
-        List<String> styles = new ArrayList<String>();
-        styles.addAll(stylesSet);
-
-        return styles;
+        Set<String> styles = new HashSet<String>();
```
pmbohm: Using HashSet means you never need to use the `uniqify` function above.
Nice!

danfruehauf: Ah yes, committed `uniqify` by accident, it should be out...

PR #56, opened 2015-10-26
Remove unused netcdf generation profiles

jkburges:
These now live in "geoserver-config" repo, e.g.:

  https://github.com/aodn/geoserver-config/blob/master/workspaces/imos/JNDI_anmn_ts/anmn_ts_timeseries_data/netcdf.xml

Pretty sure that these XML files are unused.  In any case, the anmn_ts one seems to be out of date.

It *may* be the case that the soop_sst one needs to be migrated to the geoserver-config repo, in which case, get it from [src/extension/ncdfgenerator/src/test/resources](src/extension/ncdfgenerator/src/test/resources)


danfruehauf:
Is it breaking the examples in the README?

jkburges:
> Is it breaking the examples in the README?

No, because the profiles are taken from [src/extension/ncdfgenerator/src/test/resources](src/extension/ncdfgenerator/src/test/resources).

Note I did 27a079b while I was testing this.  I think this is better, in particular, putting sensitive parameters in env vars - easier to set up for devs plus less chance of accidentally leaking (who would do that?)

jkburges:
Note: looks like [http://repo.boundlessgeo.com/main/](http://repo.boundlessgeo.com/main/) is dead - hence the travis errors.

PR #57, opened 2015-10-26
Geotools cql

jkburges:


jkburges:
Travis fails for same reason as https://github.com/aodn/geoserver-build/pull/56#issuecomment-151025120

jonescc:
I'll review

src/extension/ncdfgenerator/src/main/java/au/org/emii/ncdfgenerator/Main.java
```diff
@@ -100,7 +117,7 @@ public static void main(String[] args) {
             encoderBuilder.setTmpCreationDir(tmpDir)
                     .setDefinition(definition)
                     .setFilterExpr(cql)
-                    .setConnection(conn)
```
jonescc: Should we get rid of references to connection elsewhere?

src/extension/ncdfgenerator/src/main/java/au/org/emii/ncdfgenerator/Main.java
```diff
@@ -88,7 +92,20 @@ public static void main(String[] args) {
...
         }
 
+        JDBCDataStore dataStore = null;
+
         try {
+            Map<String, Object> dsParams = new HashMap<String, Object>();
+            dsParams.put("dbtype", "postgis");
+            dsParams.put("host", "dbprod.emii.org.au");
+            dsParams.put("port", 5432);
+            dsParams.put("database", "harvest?loginTimeout=1000&ssl=true&sslfactory=org.postgresql.ssl.NonValidatingFactory");
```
jonescc: Should we have these values here hard coded - the existing code for connection takes command line parameters

src/extension/ncdfgenerator/src/main/java/au/org/emii/ncdfgenerator/FeatureTypeFactory.java
```diff
@@ -0,0 +1,110 @@
...
+import org.geotools.jdbc.JDBCDataStore;
+import org.opengis.feature.type.*;
+import org.geotools.data.simple.SimpleFeatureIterator;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+import com.vividsolutions.jts.geom.Geometry;
+
+public class FeatureTypeFactory {
+    // Lifted from: https://github.com/geoserver/geoserver/blob/master/src/web/core/src/main/java/org/geoserver/web/data/layer/SQLViewAbstractPage.java#L404
```
jonescc: Is it worth extracting the code lifted from here into its own method - its currently just a part of the method below

src/extension/ncdfgenerator/src/main/java/au/org/emii/ncdfgenerator/Main.java
jkburges: Yeah, good spot, I've got a bit of cleaning up to do in here that I missed.

jkburges: Done: 8afe8d2

src/extension/ncdfgenerator/src/main/java/au/org/emii/ncdfgenerator/Main.java
jkburges: Oops, I'll fix that :smile:

jonescc:
Love all the deleted code. 

Pity we have to include the schema name in the sql - makes it difficult to change where data is stored.  Worth revisiting if we get time?

src/extension/ncdfgenerator/src/main/java/au/org/emii/ncdfgenerator/FeatureTypeFactory.java
jkburges: I'm not quite sure I understand what you mean here.  I think it'd be worth perhaps making the currently private geoserver code public, so that we don't have to duplicate it here, if that's what you mean?

That'll require a geoserver PR though - I don't mind attempting that, but I don't want to depend on it here (as it'll take a while no doubt).

jkburges:
> Pity we have to include the schema name in the sql - makes it difficult to change where data is stored. Worth revisiting if we get time?

Yes, it's a pity.  I'm sure it's fixable but it's going to take some digging in the geoserver code I think.

src/extension/ncdfgenerator/src/main/java/au/org/emii/ncdfgenerator/FeatureTypeFactory.java
jonescc: What I was thinking was separating out the code you have added from what you have lifted to make it clearer what was lifted (see example below).  Not sure you would just make this a public method in GeoServer - the common bits would probably need to be factored out.   Its not a biggie - just a thought I had when trying to reconcile your changes with the geoserver code - I had to work out what you had lifted and what you had added.

    public static SimpleFeatureType guessFeatureType(JDBCDataStore store, String virtualTableDef, boolean guessGeometrySrid) throws IOException {

        String vtName = "vt";

        try {
            return guessFeatureType(store, vtName, guessGeometrySrid);
       }
       finally {
           store.dropVirtualTable(vtName);
       }
    }

   
   // Lifted from: https://github.com/geoserver/geoserver/blob/master/src/web/core/src/main/java/org/geoserver/web/data/layer/SQLViewAbstractPage.java#L404
    public static SimpleFeatureType guessFeatureType(JDBCDataStore store, String vtName, boolean guessGeometrySrid) throws IOException {
        SimpleFeatureType base = store.getSchema(vtName);
        List<String> geometries = new ArrayList<String>();
        ...

jkburges: Ah ok I see.  Yeah I had thought that too actually but I was too lazy to change what I had working!  But I will, because it's a good idea!

jkburges: 9f757a8

PR #58, opened 2015-10-26
Regex restriction for uniqueValues

danfruehauf:
Allow restricting uniqueValues with a regex. By default allow only
layers with a _map suffix.

### DIFF
--- a/src/extension/layer-filters/src/main/java/au/org/emii/geoserver/extensions/filters/LayerFiltersService.java
+++ b/src/extension/layer-filters/src/main/java/au/org/emii/geoserver/extensions/filters/LayerFiltersService.java
@@ -40,6 +40,7 @@ import java.util.Set;
 public class LayerFiltersService {
 
     private Catalog catalog;
+    private List<String> uniqueValuesAllowedRegex = null;
 
     @Autowired
     private ServletContext context;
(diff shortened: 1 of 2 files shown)


danfruehauf:
Why not, why not...

danfruehauf:
Ah yes, I've not tested this yet ;)

danfruehauf:
Bring postgres down to its knees by issuing:
```
$ curl -v "http://po.aodn.org.au/geoserver/wms?request=uniqueValues&service=layerFilters&version=1.0.0&workspace=imos&layer=argo_profile_data&propertyName=project_name"
```

After this PR is merged, we can actually deny those things. By default they will be denied.

BTW, tested, works fine.

danfruehauf:
Solves the main problem that is described at https://github.com/aodn/internal_discussions/issues/130 (number 1)

jkburges:
So, if this is merged, do we need to do some cheffing to retain current portal functionality? Or not...

danfruehauf:
> So, if this is merged, do we need to do some cheffing to retain current portal functionality? Or not...

See:
https://github.com/aodn/geoserver-build/pull/58/files#diff-87eb6f9fd68dc28ebdd0c46b1ee98857R27

The default is to allow all `uniqueValues` on all layers with a `_map` suffix, on any variable and in any workspace.

danfruehauf:
So, no, no need to play with chef, unless this changes for any reason. I'll talk to IMAS though to let them know of this new behaviour, so Mark is not surprised.

danfruehauf:
@markhepburn FYI - `uniqueValues` will be allowed only on layers with a `_map` suffix by default.

danfruehauf:
Rebased after #60.

PR #61, opened 2015-10-28
Add MS SQL Server support via jtds

markhepburn:
Geoserver can support SQL Server via the gt-jdbc-sqlserver plugin, but it also needs a JDBC driver. The two options are Microsoft's own driver, free to download but not redistributable, or jTDS (http://jtds.sourceforge.net/) which is LGPL'd.  This uses jtds.

### DIFF
--- a/src/main/pom.xml
+++ b/src/main/pom.xml
@@ -65,6 +65,16 @@
             <version>${geoserver.version}</version>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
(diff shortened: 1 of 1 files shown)


danfruehauf:
Don't crucify me for that, I was initially against it, but those kind of things will just make us ensure our geoserver build is "AODN friendly".

markhepburn:
Thanks @danfruehauf :)

PR #62, opened 2015-10-28
Merge wps extensions

jkburges:


danfruehauf:
Once it passes the tests, I'm happy to merge it.

danfruehauf:
@jkburges I think you forgot to modify another pom file, look at:
https://github.com/aodn/geoserver-build/blob/master/src/main/pom.xml#L19

So it needs to include the wps mega module. You can verify that it does by `unzip -l blah.war | grep -i wps`.

jkburges:
Yeah I've just discovered that, fixing it now.

jkburges:
Aside from fixing the pom to package the new wps mega module, it turns out it's also important to have the `class` attribute in the applicationContext.xml bean definitions match the actual name of the classes as in b74a356!  Who knew?

danfruehauf:
Just noticed I will also need to amend:
https://github.com/aodn/chef-private/pull/1618

jkburges:
Ah sorry.

danfruehauf:
> Ah sorry.

Don't be, it's my PR. I'll amend it...

PR #64, opened 2015-10-30
Fix test failure due to time zone issues

jonescc:
Standardise compiler target version


src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/UrlIndexWfsHttp.java
```diff
@@ -98,10 +102,12 @@ private String cqlFilterForSameDay(String day, String timeFieldName) {
         DateTime timeEnd = timeStart.plusDays(1); // Just the next day
         LOGGER.log(Level.INFO, String.format("Returning times of day '%s'", day));
 
+        DateTimeFormatter formatter = DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
```
danfruehauf: This is ugly, but hopefully that will be out next iteration.

PR #66, opened 2015-11-04
Revert to version of json dependency that doesn't require java 8 to r…

jonescc:
…un it


danfruehauf:
I'm just quickly testing this...

danfruehauf:
@jonescc Look at #67

PR #68, opened 2015-11-05
Add postgres database support for jetty:run

jonescc:
Looked at this while trying to work out how to test my wps changes locally.  

No test data included.  May add some as a part of my testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,34 @@ Configures a GeoServer war file with the following;
 * CSV with metadata header plugin
 * Layer filter extension
 
-To build:
+### To build
 
 ```
 mvn clean install -U 
 ```
+### To setup to run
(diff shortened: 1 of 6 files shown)


danfruehauf:
I was thinking `context.xml` should be renamed to `context-sample.xml`. `context.xml` should in fact be in `.gitignore`. That way we'll prevent people from committing passwords etc by accident.

jonescc:
Well, I wasn't planning on putting any production database details in there - just to set up some basic test data.  But yeah I guess to start with people will modify it to connect to something with data.  So yes can add another step to copy context.xml

jonescc:
OK, updated

danfruehauf:
Thanks!

PR #70, opened 2015-11-06
Add timeseries sample data

jonescc:
Allows me to run gs:NetcdfOutput for imos:ts_timeseries when running mvn jetty:run


src/main/src/jetty/database/ncdfgenerator/changelog.xml
```diff
@@ -0,0 +1,83 @@
...
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd
+        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+
+    <changeSet author="talend" id="add_sample_timeseries_data">
+        <!-- From anmn_ts_timeseries harvester -->
+        <sql>
+            CREATE TABLE ts_timeseries
```
jkburges: Considering this is test data, is it worth simplifying it a bit?  I think in general that test data should be the bare minimum needed to test a particular feature, otherwise it's difficult to ascertain what's *actually* having an effect on behaviour. Also, more to maintain as the code changes.

I know that's more work, so don't have to do it - just a comment.

src/main/src/jetty/geoserver_data/demo/WCS_describeCoverage.xml
```diff
@@ -0,0 +1,19 @@
+<!-- A sample describe request.  The schema is generated automatically by  -->
```
jkburges: Do we need the `demo` directory?

src/main/src/jetty/database/ncdfgenerator/changelog.xml
jonescc: Yes its worth simplifying it .  I was just trying to get up and running to make the changes I needed so didn't do this.  I had already spent a lot of time doing this and it wasn't really a task we had identified.

jkburges:
Wouldn't it be more typical for a maven project to put test resources (the geoserver data dir in this case) under `src/test/resources`?  e.g. as with `src/extension/wps/src/test/resources`

src/main/src/jetty/geoserver_data/demo/WCS_describeCoverage.xml
jonescc: I was undecided - it gives you the example requests in GeoServer which I used a bit when trying out WPS requests.  I thought these may be useful - but no - technically not needed

jkburges:
Only a couple of comments @jonescc. 

This, along with the prior work on `mvn jetty:run`, is a great step forward for productivity :+1:

jonescc:
Re test/resources. Its not really for automated testing yet - just sample data so I could startup GeoServer to run stuff manually.

The next step could be running jetty as part of an integration test rather than standalone with real test data which would go in src/test/resources

jkburges:
:+1:

PR #78, opened 2015-11-25
Always close transactions and connections

jonescc:


jonescc:
Will include in notification before result fix

PR #80, opened 2015-11-26
Update README.md

pmbohm:


README.md
```diff
@@ -41,6 +41,10 @@ cp context-sample.xml context.xml
 ```
 cd src/main
 mvn jetty:run
+
+or something like...
```
jonescc: Perhaps change this to 

GeoServer requires extra memory above the normal maven defaults, so you will need to bump up the memory when running this command.  For example, run

    export MAVEN_OPTS='-Xms100m -Xmx512m -XX:MaxPermSize=192m'

prior to running this command or add this command to you startup scripts

To run jetty on another port you can use

     mvn jetty:run -Djetty.port=9090

PR #81, opened 2015-11-26
Include whether subset succeeded or failed in portal callback

jonescc:
Allow portal to send correct email by including whether the subset succeeded or failed in the callback


dnahodil:
This is much nicer than having to go back to GeoServer to determine the success!

PR #84, opened 2015-12-02
Reintroduce web.xml

danfruehauf:
Without that, there is no:
 * csv-with-metadata-header
 * HTTP 500 error codes

A better solution should be introduced as this can break easily on
version upgrades.

Fixes https://github.com/aodn/aodn-portal/issues/2079


danfruehauf:
@jonescc Good to go, I've experimented with it.

PR #87, opened 2015-12-03
Remove geowebcache plugin which is included by default in core geoserver

jonescc:
Geowebcache is a core plugin that we don't want - we use squid for caching and just having geowebcache installed adds minutes to the startup time.


danfruehauf:
:+1:

PR #93, opened 2015-12-03
Fix float variables in GGD

### LINKED ISSUE #91: GoGoDuck can't handle -90, -180 to 90, 180 subset
# To reproduce problem 

1. Run up latest portal/geoserver wps changes in po box
1. Modify acorn wps example to use subset TIME,2014-06-01T00:30:00.000Z,2014-06-06T02:30:00.000Z;LATITUDE,-90,90;LONGITUDE,-180,180 (which is in the format that the portal creates when no bounding box is specified)
1. use curl to submit subset for processing

# What happens

Job returns an error (not a netcdf file)

# What I expect to happen

Job finishes processing and a subsetted file is available

danfruehauf:
A bit hacky, but hopefully will be replaced so ncks is not used
and then no need to sanitize those strings.

Fixes #91.

### DIFF
--- a/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/SubsetParameters.java
+++ b/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/SubsetParameters.java
@@ -1,5 +1,6 @@
 package au.org.emii.gogoduck.worker;
 
+import java.math.BigDecimal;
 import java.util.*;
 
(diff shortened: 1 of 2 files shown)


danfruehauf:
@jonescc made me do it :cry:

danfruehauf:
Experimented with `BigDecimal`, it's OK, see e4e1101

PR #98, opened 2015-12-09
Fixes aodn/aodn-portal#2087 - use proxied url if applicable

jonescc:


danfruehauf:
I will not have time to test it, but the changes look good.

PR #101, opened 2015-12-18
Revert "Regex restriction for uniqueValues"

jonescc:
This reverts commit 5be0c0400c82e710d7f32b328cd5c4dbb315a2fa.

Refer #2080 - requires content changes/additional work.

### DIFF
--- a/src/extension/layer-filters/src/main/java/au/org/emii/geoserver/extensions/filters/LayerFiltersService.java
+++ b/src/extension/layer-filters/src/main/java/au/org/emii/geoserver/extensions/filters/LayerFiltersService.java
@@ -40,7 +40,6 @@ import java.util.Set;
 public class LayerFiltersService {
 
     private Catalog catalog;
-    private List<String> uniqueValuesAllowedRegex = null;
 
     @Autowired
     private ServletContext context;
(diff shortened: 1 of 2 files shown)


danfruehauf:
Why was this whole thing reverted? Could have been fixed with a config change - allowing `.*` in the regex.

And if that thing is not in place, I expect a **SEVERE** ticket to take its place. What at stake:
 * A few layers not following conventions not having filters vs. a completely destroyed database

I prefer one layer with no filters. If it requires additional content work, then where are the tickets for that?

This is by all means not a fix, this is reducing functionality and following an easy and dangerous path.

I'm reverting that PR and opening the required content tickets.

pblain:
@danfruehauf  - Please don't revert the PR.

danfruehauf:
@pblain I still believe there should be a proper fix for that issue. We can't just brush it off like that.

pblain:
@danfruehauf , yes there should be a proper fix - the DB vulnerability issue is a serious issue that needs solving - but unfortunately this config approach won't work for the AODN.

danfruehauf:
If we externalize it, then it will. The default can be `.*` - we can tighten it for ourselves.

pblain:
OK - if we allow .* that's fine.  But we can't dictate (yet more) non standard metadata config to AODN contributors.  Contributors need the freedom to name their own layers however they want.  It's bad enough as it is.

danfruehauf:
https://www.youtube.com/watch?v=nsJpTTIEySo

PR #102, opened 2015-12-20
Revert "Revert "Regex restriction for uniqueValues""

danfruehauf:
Reverts aodn/geoserver-build#101

Reopens https://github.com/aodn/aodn-portal/issues/2080


danfruehauf:
@jonescc Externalized properly to `filters.xml`.

danfruehauf:
See also:
https://github.com/aodn/geoserver-config/pull/310

jonescc:
@danfruehauf looks good for us.  Will this cause a problem for others using this plugin - should the code default to allowing all if no config file is found for example.

jonescc:
Actually, I can see now you've catered for that.  I'll try it out.

PR #104, opened 2016-01-04
New error message returned for stopped WPS jobs

pmbohm:


src/extension/wps/src/main/java/au/org/emii/wps/NetcdfOutputProcess.java
```diff
@@ -113,7 +112,7 @@ public FileRawData execute(
             encoder.prepare(new ZipFormatter(os));
             while (encoder.writeNext()) {
                 if (progressListener.isCanceled()) {
-                    throw new ProcessDismissedException(progressListener);
+                    throw new ProcessDismissedException("The job has been stopped due to excessive server resources required");
```
jonescc: We don't realky know why the job was cancelled - perhaps just "The job has been cancelled" would be more accurate?

jonescc:
@pmbohm - just noticed you had made changes here - perhaps add a comment next time so I get notified that you've made some changes and I should review/merge.

PR #106, opened 2016-01-07
abom proto fix tests

danfruehauf:


danfruehauf:
@jonescc Should work now.

I didn't branch from last master, and from where I branched it had the correct imports. But when travis took my work and merged it on top of current master - the imports were not there. For a moment I thought I am going insane...

danfruehauf:
@fxmzb123 FYI. Protocol changed from `IMOS:NCWMS--abom` to `IMOS:NCWMS--proto`.

PR #113, opened 2016-01-28
Fix timesteps call in ncwms controller

danfruehauf:
Return only local time, not a ISO8601 timestamp.


danfruehauf:
An amendment to #112

PR #116, opened 2016-02-12
Gridded data investigation

jonescc:
Add doco on geoserver netcdf gridded data investigation.   Is this the right place for this?


danfruehauf:
Can the commit history please be fixed here? (Assuming this is supposed to be merged)

jonescc:
@danfruehauf - commits squashed - thought it would be useful to record results of gridded data investigation and there are some associated files etc so thought this would be a good place to put it.   Other suggestions welcome

PR #122, opened 2016-03-02
Upgrade to GeoServer 2.8.2

jonescc:
Use matching gt-jdbc-sqlserver version
Use managed versions of joda-time and easymock
Use parent version for extensions so we don't have to update the version in multiple places (and fixes warnings for me in eclipse)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
     <url>http://www.emii.org.au</url>
 
     <properties>
-        <geoserver.version>2.8.0</geoserver.version>
+        <geoserver.version>2.8.2</geoserver.version>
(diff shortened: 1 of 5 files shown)


danfruehauf:
I remember there was a file we copied from upstream, just to modify a few things in it. Some XML. Remember?

danfruehauf:
Yes, it was `web.xml`.

danfruehauf:
See d159736c58f6963fa8696aa5f8a63461193c2e7b

jonescc:
Good point - but checking this, web.xml hasn't changed since 2.8.0, so should be good.

PR #123, opened 2016-03-03
Add csv point download option to gogoduck

anguss00:
Initial PR: executes on the command line.
TODO: testing as WPS module.
For review.

### DIFF
--- a/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
+++ b/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
@@ -1,0 +1,258 @@
+package au.org.emii.gogoduck.worker;
+
+import org.apache.commons.io.FilenameUtils;
+
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.commons.io.FileUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CSVConverter extends Converter {
+
+    private String ncksPath = null;
+    private String ncpdqPath = null;
+    private final String mimeType = "text/csv";
+    private final String extension = "csv";
+    private Integer lineNumber = 0;
+    private Integer attributesEnd = 0;
+    private List<String> lines = new ArrayList<String>();
+    private List<String> metadata = new ArrayList<String>();
+    private List<String> attributeMetadata = new ArrayList<String>();
+    private List<Map> attributes = new ArrayList<Map>();
+    private StringBuilder csvString = new StringBuilder();
+
+    private final String NEW_LINE = System.getProperty("line.separator");
+    private final String COMMENT = "# ";
+    private static final Logger logger = LoggerFactory.getLogger(CSVConverter.class);
+
+    public CSVConverter() {}
+
+    @Override
+    public void init() {
+        this.ncksPath = GoGoDuckConfig.ncksPath;
+        this.ncpdqPath = GoGoDuckConfig.ncpdqPath;
+    }
+
+    @Override
+    public String getMimeType() { return mimeType; }
+
+    @Override
+    public String getExtension() { return extension; }
+
+    @Override
+    public Path convert(Path outputFile) throws GoGoDuckException {
+        String baseFilename = FilenameUtils.getBaseName(String.valueOf(outputFile));
+        try {
+            File unpackedNetCDF = unpackNetCDF(outputFile.toFile());
+            Path csvPath = executeNcksAndOutputToFile(baseFilename, unpackedNetCDF.toPath());
+            Files.delete(unpackedNetCDF.toPath());
+            return csvPath;
+        } catch (Exception e) {
+            throw new GoGoDuckException(String.format("Could not convert output to csv: '%s'", e.getMessage()));
+        }
+    }
+
+    private File unpackNetCDF(File netCDFInputFile) throws Exception {
+        List<String> ncpdqCommand = new ArrayList<>();
+        File unpackedNetCDF = File.createTempFile("tmp", ".nc");
+
+        ncpdqCommand.add(ncpdqPath);
+        ncpdqCommand.add("-U");
+        ncpdqCommand.add("-O");
+        ncpdqCommand.add(netCDFInputFile.toString());
+        ncpdqCommand.add(unpackedNetCDF.getPath());
+
+        GoGoDuck.execute(ncpdqCommand);
+        return unpackedNetCDF;
+    }
+
+    private Path executeNcksAndOutputToFile(String baseFilename, Path tempFile) throws IOException, InterruptedException {
+        List<String> ncksCommand = new ArrayList<>();
+        ncksCommand.add(ncksPath);
+        ncksCommand.add("--no_nm_prn");
+        ncksCommand.add("-a");
+        ncksCommand.add(tempFile.toString());
+
+        logger.info(ncksCommand.toString());
+        
+        ProcessBuilder processor = new ProcessBuilder(ncksCommand);
+        processor.redirectError(ProcessBuilder.Redirect.PIPE);
+
+        final Process process = processor.start();
+        InputStream is = process.getInputStream();
+        InputStreamReader isr = new InputStreamReader(is);
+        BufferedReader br = new BufferedReader(isr);
+
+        InputStreamReader errorStream = new InputStreamReader(process.getErrorStream());
+        BufferedReader bufferedErrorStream = new BufferedReader(errorStream);
+        
+        StringBuilder builder = new StringBuilder();
+        String line;
+        String errLine = null;
+        while ((line = br.readLine()) != null || (errLine = bufferedErrorStream.readLine()) != null) {
+            if (line != null) {
+                builder.append(line + System.getProperty("line.separator"));
+            }
+            if (errLine != null) {
+                logger.error(errLine);
+            }
+        }
+
+        try {
+            process.waitFor();
+        } catch (InterruptedException e) {
+            logger.error(String.format("Interrupted: '%s'", e.getMessage()));
+            throw e;
+        }
+
+        String rawOutput = builder.toString();
+        String outputCSV = parseNcks(rawOutput);
+        String csvFilename = String.format("%s.%s", baseFilename, extension);
+        File csvFile = new File(csvFilename);
+        FileUtils.writeStringToFile(csvFile, outputCSV);
+
+        return csvFile.toPath();
+    }
+
+    private String parseNcks(String input) {
+        InputStream is = new ByteArrayInputStream(input.getBytes());
+
+        try (BufferedReader br = new BufferedReader(new InputStreamReader(is))) {
+            String line = br.readLine();
+
+            while (line != null) {
+                extractGlobalMetadata(line);
+                extractAttributes(line);
+                lines.add(line);
+
+                line = br.readLine();
+                lineNumber++;
+            }
+
+            for (Map attr : attributes) {
+                extractDetailedAttributes(attr);
+            }
+
+            writeMetadata();
+            writeDataBlocks();
+        }
+        catch (Exception e) {
+            logger.error(e.toString());
+        }
+
+        return csvString.toString();
+    }
+
+    private void extractGlobalMetadata(String aText) {
+        Pattern pattern = Pattern.compile("^Global", Pattern.COMMENTS);
+        Matcher matcher = pattern.matcher(aText);
+
+        if (matcher.find()) {
+            metadata.add(String.format(COMMENT + "%s", aText.split(":", 2)[1].trim()));
+        }
+    }
+
+    private void extractAttributes(String aText) {
+        Pattern pattern = Pattern.compile("\\:\\stype", Pattern.COMMENTS);
+        Matcher matcher = pattern.matcher(aText);
+
+        while (matcher.find()) {
+            Map<String, String> attr = new HashMap<String, String>();
+            attr.put("name", aText.split(":", 2)[0].trim());
+            attr.put("lineNumber", lineNumber.toString());
+
+            attributes.add(attr);
+        }
+    }
+
+    private void extractDetailedAttributes(Map attr) {
+        String lNum = attr.get("lineNumber").toString();
+        Integer lineNumber = Integer.parseInt(lNum);
+        String line = lines.get(lineNumber);
+        attributeMetadata.add(COMMENT);
+
+        while (line.length() > 0) {
+            setHeaders(attr,line);
+            attributeMetadata.add(COMMENT + line);
+            line = lines.get(lineNumber);
+            lineNumber ++;
+            setEndOfVariableBlock(lineNumber);
+        }
+    }
+
+    private void setHeaders(Map attr, String line) {
+        Pattern pattern = Pattern.compile("\\sdimension\\s[0-9]:", Pattern.COMMENTS);
+        Matcher matcher = pattern.matcher(line);
+        Integer dimensionCount = 1;
+
+        while (matcher.find()) {
+            String dimensionName = line.split(":", 2)[1].split(",", 2)[0].trim();
+
+            if (attr.get("dimensions") == null) {
+                attr.put("dimensions", dimensionName);
+            } else {
+                attr.put("dimensions", attr.get("dimensions") + "," + dimensionName);
+            }
+
+            attr.put("dimension_count", dimensionCount);
+            dimensionCount ++;
+        }
+    }
+
+    private void writeMetadata() {
+        csvAppender("# METADATA:");
+
+        for (String m : metadata) {
+            csvAppender(m);
+        }
+
+        for (String a : attributeMetadata) {
+            csvAppender(a);
+        }
+        csvAppender(NEW_LINE);
+    }
+
+    private void writeDataBlockHeaders(Map attr) {
+        String header = attr.get("dimensions") + "," + attr.get("name");
+        String dimensions = attr.get("dimensions").toString().replace(",", "");
+
+        if (dimensions.equals(attr.get("name").toString())) {
+            header = attr.get("name").toString();
+        }
+
+        csvAppender(header);
+    }
+
+    private void writeDataBlocks() {
+        for (Map attr : attributes) {
+            writeDataBlockHeaders(attr);
+            String line = lines.get(attributesEnd );
+
+            while (line.length() > 0) {
+                line = lines.get(attributesEnd);
+                attributesEnd++;
+
+                csvAppender(line.replace(" ", ","));
+            }
+        }
+    }
+
+    private void setEndOfVariableBlock(Integer lineNumber) {
+        if (lineNumber > attributesEnd) {
+            attributesEnd = lineNumber;
+        }
+    }
+
+    private void csvAppender(String line) {
+        csvString.append(line + NEW_LINE);
+    }
+}
--- a/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/Converter.java
+++ b/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/Converter.java
@@ -1,0 +1,36 @@
+package au.org.emii.gogoduck.worker;
+
+import java.nio.file.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class Converter {
+
+    public abstract void init();
+
+    public abstract Path convert(Path outputFile) throws GoGoDuckException;
+
+    public abstract String getMimeType();
+
+    public abstract String getExtension();
+
+    public static Converter newInstance(String filter) {
+        Logger logger = LoggerFactory.getLogger(Converter.class);
+        String thisPackage = Converter.class.getPackage().getName();
+        String classToInstantiate = String.format("%sConverter", filter.toUpperCase());
+
+        logger.debug(String.format("Trying class '%s.%s'", thisPackage, classToInstantiate));
+        try {
+            Class classz = Class.forName(String.format("%s.%s", thisPackage, classToInstantiate));
+            Converter converter = (Converter) classz.newInstance();
+            converter.init();
+            logger.info(String.format("Using class '%s.%s'", thisPackage, classToInstantiate));
+            return converter;
+        }
+        catch (Exception e) {
+            logger.debug(String.format("Could not find class for '%s.%s'", thisPackage, classToInstantiate));
+        }
+
+        throw new GoGoDuckException(String.format("Error initializing class for filter '%s'", filter));
+    }
+}
--- a/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
+++ b/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
@@ -1,5 +1,12 @@
 package au.org.emii.gogoduck.worker;
 
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.FilenameUtils;
+import org.geoserver.catalog.Catalog;
+import org.opengis.util.ProgressListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.*;
 import java.net.URI;
 import java.net.URL;
@@ -7,16 +14,6 @@ import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
-
-import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.FilenameUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.opengis.util.ProgressListener;
-
-import org.geoserver.catalog.Catalog;
-import java.io.IOException;
 import java.util.*;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -30,36 +27,36 @@ public class GoGoDuck {
     // here merely to have a value to satisfy awaitTermination of the JAVA API
     private static final int MAX_EXECUTION_TIME_DAYS = 365;
 
-    public static final String ncksPath = "/usr/bin/ncks";
-    public static final String ncrcatPath = "/usr/bin/ncrcat";
-    public static final String ncdpqPath = "/usr/bin/ncpdq";
-
     private IndexReader indexReader = null;
     private final String profile;
     private final String subset;
-    private final Path outputFile;
+    private final List<Converter> converters;
+    private Path outputFile;
     private final Integer limit;
     private Path baseTmpDir;
     private int threadCount = 1;
     private UserLog userLog = null;
     private ProgressListener progressListener = null;
+    private String mimeType = "application/x-netcdf";
+    private String extension = "nc";
 
-    protected GoGoDuck(String profile, String subset, String outputFile, Integer limit) {
+    protected GoGoDuck(String profile, String subset, String outputFile, List<Converter> converters, Integer limit) {
         this.profile = profile;
         this.subset = subset;
         this.outputFile = new File(outputFile).toPath();
+        this.converters = converters;
         this.limit = limit;
         this.baseTmpDir = new File(System.getProperty("java.io.tmpdir")).toPath();
         this.userLog = new UserLog();
     }
 
-    public GoGoDuck(String geoserverUrl, String profile, String subset, String outputFile, Integer limit) {
-        this(profile, subset, outputFile, limit);
+    public GoGoDuck(String geoserverUrl, String profile, String subset, String outputFile, List<Converter> converters, Integer limit) {
+        this(profile, subset, outputFile, converters, limit);
         this.indexReader = new HttpIndexReader(this.userLog, geoserverUrl);
     }
 
-    public GoGoDuck(Catalog catalog, String profile, String subset, String outputFile, Integer limit) {
-        this(profile, subset, outputFile, limit);
+    public GoGoDuck(Catalog catalog, String profile, String subset, String outputFile, List<Converter> converters, Integer limit) {
+        this(profile, subset, outputFile, converters, limit);
         this.indexReader = new FeatureSourceIndexReader(this.userLog, catalog);
     }
 
@@ -94,7 +91,7 @@ public class GoGoDuck {
             throw new GoGoDuckException("Job cancelled");
     }
 
-    public void run() {
+    public Path run() {
         GoGoDuckModule module = GoGoDuckModule.newInstance(profile, indexReader, subset, userLog);
 
         Path tmpDir = null;
@@ -115,6 +112,12 @@ public class GoGoDuck {
             throwIfCancelled();
             updateMetadata(module, outputFile);
             throwIfCancelled();
+            for (Converter converter : converters) {
+                outputFile = converter.convert(outputFile);
+                mimeType = converter.getMimeType();
+                extension = converter.getExtension();
+                throwIfCancelled();
+            }
             userLog.log("Your aggregation was successful!");
         }
         catch (Exception e) {
@@ -126,6 +129,7 @@ public class GoGoDuck {
         finally {
             cleanTmpDir(tmpDir);
             userLog.close();
+            return outputFile;
         }
     }
 
@@ -239,7 +243,7 @@ public class GoGoDuck {
             File tmpFile = File.createTempFile("tmp", ".nc");
 
             List<String> command = new ArrayList<String>();
-            command.add(ncksPath);
+            command.add(GoGoDuckConfig.ncksPath);
             command.add("-a");
             command.add("-4");
             command.add("-O");
@@ -337,7 +341,7 @@ public class GoGoDuck {
 
     private static void aggregateNcrcat(Path tmpDir, Path outputFile) throws GoGoDuckException {
         List<String> command = new ArrayList<String>();
-        command.add(ncrcatPath);
+        command.add(GoGoDuckConfig.ncrcatPath);
         command.add("-D2");
         command.add("-4");
         command.add("-h");
@@ -420,4 +424,8 @@ public class GoGoDuck {
 
         return process.exitValue();
     }
+
+    public String getMimeType() { return mimeType; }
+
+    public String getExtension() { return extension; }
 }
--- a/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckConfig.java
+++ b/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckConfig.java
@@ -1,0 +1,15 @@
+package au.org.emii.gogoduck.worker;
+
+public class GoGoDuckConfig {
+
+    public static final String CONFIG_FILE = "wps/gogoduck.xml";
+    public static final String FILE_LIMIT_KEY = "/gogoduck/fileLimit";
+    public static final String THREAD_COUNT_KEY = "/gogoduck/threadCount";
+    public static final String FILE_LIMIT_DEFAULT = "10";
+    public static final String THREAD_COUNT_DEFAULT = "0";
+
+
+    public static final String ncksPath = "/usr/bin/ncks";
+    public static final String ncpdqPath = "/usr/bin/ncpdq";
+    public static final String ncrcatPath = "/usr/bin/ncrcat";
+}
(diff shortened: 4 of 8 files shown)


src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/Main.java
```diff
@@ -55,6 +58,7 @@ public static void main(String[] args) {
         String limit = cmd.getOptionValue("l", "100");
         String threadCount = cmd.getOptionValue("n", "1");
         String outputFile = cmd.getOptionValue("o");
+        String filter = cmd.getOptionValue("f");
```
danfruehauf: Filter or format? Look at https://github.com/aodn/geoserver-build/pull/123/files#diff-49c51c5d74954153f8662c47021a5764R40

anguss00: Good point

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/NcksToCsv.java
```diff
@@ -0,0 +1,176 @@
...
+    public static List<String> lines = new ArrayList<String>();
+    public static List<String> metadata = new ArrayList<String>();
+    public static List<String> attributeMetadata = new ArrayList<String>();
+    public static List<Map> attributes = new ArrayList<Map>();
+    public static StringBuilder csvString = new StringBuilder();
+
+    private static final String fNEW_LINE = System.getProperty("line.separator");
+    private static final String fCOMMENT = "# ";
+
+    public static void main (String... aArguments) {
```
danfruehauf: Tidy up this file. No new lines after function declarations, and also no space after function name before brackets.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckConfig.java
```diff
@@ -0,0 +1,15 @@
+package au.org.emii.gogoduck.worker;
+
+public class GoGoDuckConfig {
```
danfruehauf: This is good.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -30,36 +28,36 @@
...
     private final Integer limit;
     private Path baseTmpDir;
     private int threadCount = 1;
     private UserLog userLog = null;
     private ProgressListener progressListener = null;
+    private String mimeType = "application/x-netcdf";
+    private String extension = "nc";
 
-    protected GoGoDuck(String profile, String subset, String outputFile, Integer limit) {
+    protected GoGoDuck(String profile, String subset, String outputFile, Converter converter, Integer limit) {
```
danfruehauf: Should take an array of converters ideally and you should iterate on them. You can say we support just one now, but should still take an array. The necessary plumbing should probably be in place.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/Converter.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.apache.commons.lang.WordUtils;
+
+import java.nio.file.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class Converter {
+
+    public abstract void init();
```
danfruehauf: Space those function declarations please...

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CsvConverter.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.io.FileUtils;
+
+public class CsvConverter extends Converter {
```
danfruehauf: Lets call it CSVConverter. It's an acronym :)

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -115,6 +113,13 @@ public void run() {
             throwIfCancelled();
             updateMetadata(module, outputFile);
             throwIfCancelled();
+            if (converter != null) {
```
danfruehauf: Should be a for loop which iterates on all the converters, rather than a `converter != null` thing...

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/NcksToCsv.java
```diff
@@ -0,0 +1,176 @@
...
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class NcksToCsv {
```
danfruehauf: I don't understand why there is another class. This implementation belongs to `CSVConverter`. Unless you *really* want it in this class.

src/extension/wps/src/main/java/au/org/emii/wps/GoGoDuckProcess.java
```diff
@@ -75,20 +68,24 @@ public FileRawData execute(
                 throw new ProcessException("fileLimit set to 0 or below, job will not run");
             }
 
-            final File outputFile = getResourceManager().getOutputResource(
+            if (filter != null && !filter.isEmpty()) {
```
danfruehauf: So this should be a comma separated list of filters. Make an array out of it, and pass it to GGD.

danfruehauf:
All up pretty good Mr. Sebulba. What I'm missing as well is the README modifications.

So ideally, I'd like it to be refactored to pass an array of converters, which will require a bit more work, but will get rid of ifs and elses which we don't like.

anguss00:
Yup good stuff. I'll push up the cleanup stuff now, and the rest when I get back

danfruehauf:
:+1: Good job though. I hope you got to learn some of the inner workings of GGD + how to use reflections.

anguss00:
Ok changes up @danfruehauf

src/extension/wps/src/main/java/au/org/emii/wps/GoGoDuckProcess.java
```diff
@@ -75,20 +68,29 @@ public FileRawData execute(
...
+                List<String> filterList = new ArrayList<>(Arrays.asList(filter.split("\\s*,\\s*")));
```
danfruehauf: Why not just split on `","`?

anguss00: Yeah, we can do that. Save the regex

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/NcksToCsv.java
anguss00: Yup, sorry. Missed this one. Yup, let's put it in CSVConverter. It's cleaner

anguss00:
Ok @danfruehauf, got a refactor up

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+        outputCommand.add(ncpdqPath + " -U " + outputFile.toString() + " tmp");
+        outputCommand.add("mv tmp " + outputFile.toString());
+        outputCommand.add(ncksPath + " --no_nm_prn -a " + outputFile.toString() + " > " + txtFilename);
+
+        try {
+            executeCommands(outputCommand, outputFile);
+            String outputFileAsCsvString = parseNcks(txtFilename);
+            Path tmpPath = Paths.get(txtFilename);
+            Files.delete(tmpPath);
+            String csvFilename = baseFilename + "." + extension;
```
danfruehauf: I'd use `String.format` here.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+    private void extractDetailedAttributes(Map attr) {
+        String lNum = attr.get("lineNumber").toString();
+        Integer lineNumber = Integer.parseInt(lNum);
+
+        String line = lines.get(lineNumber);
+
+        attributeMetadata.add(fCOMMENT);
+
+        while (line.length() > 0) {
+
```
danfruehauf: Empty line - remove.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+        for (String command : commandList) {
+            printWriter.println(command);
+        }
+
+        printWriter.close();
+
+        return script;
+    }
+
+    private String parseNcks(String... aArguments) {
```
danfruehauf: Too many spaces in that function.

danfruehauf: What is `String...`? `aArguments`?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+        catch (Exception e) {
+            throw new GoGoDuckException(String.format("Could not convert output to csv: '%s'", e.getMessage()));
+        }
+    }
+
+    private void executeCommands(List<String> commandList, Path outputFile) throws IOException {
+        File tmpScript = createCommandScript(commandList, outputFile);
+
+        try {
+            ProcessBuilder processor = new ProcessBuilder("bash", tmpScript.toString());
```
danfruehauf: `"bash"`? - absolute no. You want to execute `ncks`, not bash.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+            tmpScript.delete();
+        }
+    }
+
+    private File createCommandScript(List<String> commandList, Path outputFile) throws IOException {
+        File script = File.createTempFile("script", null);
+        Writer writer = new OutputStreamWriter(new FileOutputStream(script));
+        PrintWriter printWriter = new PrintWriter(writer);
+
+        printWriter.println("#!/bin/bash");
```
danfruehauf: WTF?!?!?!

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    private void writeDataBlocks() {
+        for (Map attr : attributes) {
+
+            writeDataBlockHeaders(attr);
+            String line = lines.get(attributesEnd );
+
+            while (line.length() > 0) {
+
```
danfruehauf: Empty line.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+        if (dimensions.equals(attr.get("name").toString())) {
+            header = attr.get("name").toString();
+        }
+
+        csvAppender(header);
+    }
+
+    private void writeDataBlocks() {
+        for (Map attr : attributes) {
+
```
danfruehauf: Empty line.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+        csvAppender("# METADATA:");
+        for (String m : metadata) {
+            csvAppender(m);
+        }
+
+        for (String a : attributeMetadata) {
+            csvAppender(a);
+        }
+        csvAppender(fNEW_LINE);
+
```
danfruehauf: Empty line.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+    @Override
+    public String getExtension() { return extension; }
+
+    @Override
+    public Path convert(Path outputFile) throws GoGoDuckException {
+        List<String> outputCommand = new ArrayList<String>();
+        String baseFilename = FilenameUtils.getBaseName(String.valueOf(outputFile));
+        String txtFilename = baseFilename + ".txt";
+        outputCommand.add(ncpdqPath + " -U " + outputFile.toString() + " tmp");
+        outputCommand.add("mv tmp " + outputFile.toString());
```
danfruehauf: ?!

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.commons.io.FileUtils;
+
+public class CSVConverter extends Converter {
+
+    private String ncksPath = null;
```
danfruehauf: You can go:
```
private final String ncksPath = GoGoDuckConfig.ncksPath;
private final String ncpdqPath = GoGoDuckConfig.ncpdqPath;
```

Then get rid of those in the ctor.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+
+    @Override
+    public String getExtension() { return extension; }
+
+    @Override
+    public Path convert(Path outputFile) throws GoGoDuckException {
+        List<String> outputCommand = new ArrayList<String>();
+        String baseFilename = FilenameUtils.getBaseName(String.valueOf(outputFile));
+        String txtFilename = baseFilename + ".txt";
+        outputCommand.add(ncpdqPath + " -U " + outputFile.toString() + " tmp");
```
danfruehauf: This is not a good way to create a temporary file. Use `File.createTempFile`

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -115,6 +113,13 @@ public void run() {
             throwIfCancelled();
             updateMetadata(module, outputFile);
             throwIfCancelled();
+            for (Converter converter : converters) {
+                Path outputPath = converter.convert(outputFile);
```
danfruehauf: Why not just:
```
outputFile = converter.convert(outputFile);
```

danfruehauf: So Ideally you want to:
 * Create a temporary file
 * Call `tmpFile = converter.convert(outputFile)`
 * Move `tmpFile` => `outputFile`
 * Keep going with the loop

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+                lineNumber ++;
+            }
+
+            for (Map attr : attributes) {
+                extractDetailedAttributes(attr);
+            }
+
+            writeMetadata();
+            writeDataBlocks();
+
```
danfruehauf: Empty line.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,243 @@
...
+    @Override
+    public String getMimeType() { return mimeType; }
+
+    @Override
+    public String getExtension() { return extension; }
+
+    @Override
+    public Path convert(Path outputFile) throws GoGoDuckException {
+        List<String> outputCommand = new ArrayList<String>();
+        String baseFilename = FilenameUtils.getBaseName(String.valueOf(outputFile));
```
danfruehauf: You shouldn't care about `basename` - create temporary files in a secure way. This file takes a file and returns another. The file it returns should be a newly created temporary. It must not touch `outputFile`.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
anguss00: I'm not sure how this is different from what I'm doing here... I create a new Path pointing to the csv file that converter generates, and then I point outputFile (which is a path) to the new path

danfruehauf: You're right, I don't think it's different. Unless you care about the filename.

Either way, you can save a line.

anguss00:
Ok latest changes up @danfruehauf

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+            throw new GoGoDuckException(String.format("Could not convert output to csv: '%s'", e.getMessage()));
+        }
+    }
+
+    private Path executeNcks(List<String> commandList, String baseFilename, Path tempFile) throws IOException, InterruptedException {
+        logger.info(commandList.toString());
+
+        File errorTmpFile = File.createTempFile("error", ".txt");
+        ProcessBuilder processor = new ProcessBuilder(commandList);
+        processor.redirectError(errorTmpFile);
```
danfruehauf: No need for an error file. Best if you dump it to the log...

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+
+        final Process process = processor.start();
+        InputStream is = process.getInputStream();
+        InputStreamReader isr = new InputStreamReader(is);
+        BufferedReader br = new BufferedReader(isr);
+        StringBuilder builder = new StringBuilder();
+
+        String line;
+        while ((line = br.readLine()) != null) {
+            builder.append(line + "\n");
```
danfruehauf: Probably a better way to obtain the newline separator for the system (instead of using `\n`).

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        try {
+            process.waitFor();
+
```
danfruehauf: Newline?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+        } catch (InterruptedException e) {
+            logger.error(String.format("Interrupted: '%s'", e.getMessage()));
+            throw e;
+        }
+
+        String rawOutput = builder.toString();
+        String outputCSV = parseNcks(rawOutput);
+        Files.delete(errorTmpFile.toPath());
+        Files.delete(tempFile);
+        String csvFilename = baseFilename + "." + extension;
```
danfruehauf: `String.format()`

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+        attributeMetadata.add(COMMENT);
+
+        while (line.length() > 0) {
+            setHeaders(attr,line);
+            attributeMetadata.add(COMMENT + line);
+            line = lines.get(lineNumber);
+            lineNumber ++;
+            setEndOfVariableBlock(lineNumber);
+        }
+
```
danfruehauf: New line

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+        for (String m : metadata) {
+            csvAppender(m);
+        }
+
+        for (String a : attributeMetadata) {
+            csvAppender(a);
+        }
+        csvAppender(NEW_LINE);
+
```
danfruehauf: Newline

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+    @Override
+    public String getExtension() { return extension; }
+
+    @Override
+    public Path convert(Path outputFile) throws GoGoDuckException {
+        List<String> ncpdqCommand = new ArrayList<>();
+        List<String> ncksCommand = new ArrayList<>();
+        String baseFilename = FilenameUtils.getBaseName(String.valueOf(outputFile));
+        try {
+            File tmpNcdf = File.createTempFile("tmp", ".nc");
```
danfruehauf: Why not just `File.createTempFile()`?

danfruehauf: And you can call it `File unpackedNetCDF`.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+
+    @Override
+    public String getMimeType() { return mimeType; }
+
+    @Override
+    public String getExtension() { return extension; }
+
+    @Override
+    public Path convert(Path outputFile) throws GoGoDuckException {
+        List<String> ncpdqCommand = new ArrayList<>();
```
danfruehauf: Any reason you keep those declarations outside the block?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+            ncpdqCommand.add(tmpNcdf.getPath());
+            GoGoDuck.execute(ncpdqCommand);
+
+            // Output temp csv and store values
+            ncksCommand.add(ncksPath);
+            ncksCommand.add("--no_nm_prn");
+            ncksCommand.add("-a");
+            ncksCommand.add(tmpNcdf.getPath());
+
+            return executeNcks(ncksCommand, baseFilename, tmpNcdf.toPath());
```
danfruehauf: You forget to delete `tmpNcdf` (`unpackedNetCDF`).

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
anguss00: How do we do that? redirectError only allows you to redirect to a file?

danfruehauf: I think you can redirect to a buffer. Much like what happens with `stdout`.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
anguss00: Because that's not a valid method, you need to have a filename and extension

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
danfruehauf: So you clean it in `executeNcks`, but you probably want to clean it up here, in the same scope. You don't want to rely on other functions to clean up the temporary files that you create in *this* scope.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
danfruehauf: Yes, my bad. Bloody Java.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+        } catch (Exception e) {
+            throw new GoGoDuckException(String.format("Could not convert output to csv: '%s'", e.getMessage()));
+        }
+    }
+
+    private Path executeNcks(List<String> commandList, String baseFilename, Path tempFile) throws IOException, InterruptedException {
```
danfruehauf: You can make this function return an `InputStream`, then do the writing into a file in the calling function. Or call it `executeNcksAndOutputToFile` - because this is what it does. It doesn't just execute ncks, it does a bit more...

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
anguss00: Indeed, I don't like it either

danfruehauf: So maybe move those things to a different functions, such as:
```
File unpackedNetDF = unpackNetCDF(File netcdfFile);
```

That function is basically the few lines below calling `ncpdq`.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+            File tmpNcdf = File.createTempFile("tmp", ".nc");
+            // Unpack packed CSV values and store in temp csv
+            ncpdqCommand.add(ncpdqPath);
+            ncpdqCommand.add("-U");
+            ncpdqCommand.add("-O");
+            ncpdqCommand.add(outputFile.toString());
+            ncpdqCommand.add(tmpNcdf.getPath());
+            GoGoDuck.execute(ncpdqCommand);
+
+            // Output temp csv and store values
```
danfruehauf: And those can go to the `executeNcks` function, because the logic on how to run `ncks` belongs more to a function called `executeNcks`.

In `convert()` you should probably leave only the high level logic. Something like:
```
File unpackedNetDF = unpackNetCDF(netcdfFile);

InputStream ncksOutput = executeNcks(netcdfFile);
Files.delete(unpackedNetCDF);

File output = File.createTempFile("tmp", ".nc");
<READ ncksOutput into a output>

return output;
```

`netCDFFile` or `netcdfFile`, I'm not convinced, something should work as long as it is consistent.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
anguss00: Not sure you can

danfruehauf: You might have to use a different method, but it is probably possible to consume stderr without redirecting to a file.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
anguss00: OK, but then you have a potential race condition between the tempFile being used to execute a command and being deleted. Because it is used in the ncks command

danfruehauf: No race condition, you need to:
 * Store the result from `executeNcks`
 * Delete your temporary file
 * Return the result

It's tedious, but those practices will prevent you from forgetting things lying around. But look at my other comment about how to structure this function and keeping only high level logic in it.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
danfruehauf: http://docs.oracle.com/javase/7/docs/api/java/lang/ProcessBuilder.html#redirectError%28java.lang.ProcessBuilder.Redirect%29

anguss00: So you're saying Redirect.PIPE?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
danfruehauf: I'll explain again what contributes to "secure" and "safe" programming. Imagine you have a block like:
```
{
<ALLOCATE_MEMORY>
...
Do something
...
<DEALLOCATE MEMORY>
}
```

If I ask you whether memory is being deallocated, you gonna answer "YES" immediately. Because you allocate and then deallocate.

If you have something like:
```
{
<ALLOCATE_MEMORY>
...
Call a function that may deallocate allocated memory
...
}
```

If I ask the same question again, your answer may be delayed. This, together with the fact that someone editing the function that you are calling and not understanding exactly the flow might remove the deallocation of memory. And then bam - memory never gets deallocated.

It's a bit of programming style. If I do a `malloc()`, or `mktemp()`, I instantly call `free()` or `unlink()` respectively and *only then* I start populating the lines between those.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
danfruehauf: I'm not entirely sure, it's for you to discover. I ain't java guru...

anguss00:
Next round @danfruehauf

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+    private Path executeNcksAndOutputToFile(String baseFilename, Path tempFile) throws IOException, InterruptedException {
+        List<String> ncksCommand = new ArrayList<>();
+        ncksCommand.add(ncksPath);
+        ncksCommand.add("--no_nm_prn");
+        ncksCommand.add("-a");
+        ncksCommand.add(tempFile.toString());
+
+        logger.info(ncksCommand.toString());
+
+        //File errorTmpFile = File.createTempFile("error", ".txt");
```
danfruehauf: Forgot this?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+            process.waitFor();
+        } catch (InterruptedException e) {
+            logger.error(String.format("Interrupted: '%s'", e.getMessage()));
+            logger.error(process.getErrorStream().toString());
+            throw e;
+        }
+
+        String rawOutput = builder.toString();
+        String outputCSV = parseNcks(rawOutput);
+        //Files.delete(errorTmpFile.toPath());
```
danfruehauf: Umn?

anguss00: Already done

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,252 @@
...
+        String line;
+        while ((line = br.readLine()) != null) {
+            builder.append(line + System.getProperty("line.separator"));
+        }
+
+        try {
+            process.waitFor();
+        } catch (InterruptedException e) {
+            logger.error(String.format("Interrupted: '%s'", e.getMessage()));
+            logger.error(process.getErrorStream().toString());
```
danfruehauf: I think you should consume the error stream regardless of catching exceptions. Regarding that @jonescc is a java guru. Can be much more help than me.

jonescc: Last time I had to do something like this I used apache commons exec.  ensuring both stdout and stderr are consumed can be tricky unless your streaming them to a file and the solutions I found used apache commons exec PumpStreamHandler

danfruehauf: Just discussed that. @anguss00 can use another buffered reader to consume stderr.

Tricky executing stuff in java :)

jonescc: I had to do pretty much the same thing in python - its tricky either way - don't write something yourself its tricky

anguss00:
Changes up @danfruehauf

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,257 @@
...
+        InputStreamReader isr = new InputStreamReader(is);
+        BufferedReader br = new BufferedReader(isr);
+
+        InputStreamReader errorStream = new InputStreamReader(process.getErrorStream());
+        BufferedReader bufferedErrorStream = new BufferedReader(errorStream);
+        
+        StringBuilder builder = new StringBuilder();
+
+        String line;
+        while ((line = br.readLine()) != null) {
```
danfruehauf: This should be after you called `process.waitFor()` I believe.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
```diff
@@ -0,0 +1,257 @@
...
+            builder.append(line + System.getProperty("line.separator"));
+        }
+
+        try {
+            process.waitFor();
+        } catch (InterruptedException e) {
+            logger.error(String.format("Interrupted: '%s'", e.getMessage()));
+            String errLine;
+            while ((errLine = bufferedErrorStream.readLine()) != null) {
```
danfruehauf: And those lines must be called regardless of catching an exception.

So you should consume both stdout and stderr in a finally block I think.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
anguss00: Why's that? Here I was following the example execute method in GoGoDuck

danfruehauf: Yep, you're right, because `processor.start()` will actually start the process. Still, consuming stderr should be done regardless of catching exceptions and probably just here, below the consumption of stdout.

anguss00: OK, do you think that's better than consuming in the finally block?

danfruehauf: It's not about what I think is better, it's about what is correct :)

And you should probably consume both of them at the same time, something like:
```
while (br.available() > 0 || bufferedErrorStream.available() > 0) {
    String line;
    while ((line = br.readLine()) != null) {
        builder.append(line + System.getProperty("line.separator"));
    }
    while ((errLine = bufferedErrorStream.readLine()) != null) {
        logger.error(errLine);
    }
}
```

**UNTESTED CODE**

anguss00: What I mean is, consuming stderr in the finally block will have exactly the same result as consuming it at the same time as stdout... But as shown above it may be cleaner (therefore "better")

anguss00:
Ok, got another set of changes @danfruehauf

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/CSVConverter.java
danfruehauf: Lets try this?
```
String line;
String errLine;
while ((line = br.readLine()) != null || (errLine = bufferedErrorStream.readLine()) != null) {
    if (line != null) {
        builder.append(line + System.getProperty("line.separator"));
    }
    if (errLine != null) {
        logger.error(errLine);
    }
}
```

anguss00:
Doneski's @danfruehauf

anguss00:
borat-quote-great-success

PR #126, opened 2016-03-23
WIP GetCapabilities for abomination controller

danfruehauf:
**DO NOT MERGE**

Since I know the abomination controller will stay here for a while, I have started to implement `GetCapabilities` for it. It is not ready, but it is not far from being *ready*. This work can save some time when actually getting to implement it - if at all.


jonescc:
Unlikely to proceed with this.

PR #133, opened 2016-06-13
Include TIME selection in thredds ncwms requests

jonescc:
Enables support for datasets containing multiple times such as CARS

### DIFF
--- a/src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/Ncwms.java
+++ b/src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/Ncwms.java
@@ -1,6 +1,7 @@
 package au.org.emii.geoserver.wms;
 
 import net.sf.json.JSONObject;
+
 import org.apache.commons.io.IOUtils;
 import org.dom4j.Document;
 import org.dom4j.DocumentException;
@@ -14,9 +15,10 @@ import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-
+import java.util.regex.Pattern;
 import java.net.URL;
 import java.util.*;
+
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -31,16 +33,19 @@ public class Ncwms {
          <wfsServer>http://localhost:8080/geoserver/ows</wfsServer>
          <urlSubstitution key="/mnt/imos-t3/IMOS/opendap/">http://thredds-1-aws-syd.aodn.org.au/thredds/wms/IMOS/</urlSubstitution>
          <urlSubstitution key="^/IMOS/">http://thredds-1-aws-syd.aodn.org.au/thredds/wms/IMOS/</urlSubstitution>
+         <collectionsWithTimeMismatch>^imos:srs.*</collectionsWithTimeMismatch>
        </ncwms>
     */
 
     private final Map<String, String> urlSubstitutions;
+    private final List<String> collectionsWithTimeMismatchRegExs;
 
     private final UrlIndexInterface urlIndexInterface;
 
     public Ncwms(UrlIndexInterface urlIndexInterface, NcwmsConfig ncwmsConfig) {
         this.urlIndexInterface = urlIndexInterface;
         urlSubstitutions = ncwmsConfig.getConfigMap("/ncwms/urlSubstitution");
+        collectionsWithTimeMismatchRegExs = ncwmsConfig.getConfigList("/ncwms/collectionsWithTimeMismatch");
 
         for (Map.Entry<String, String> entry : urlSubstitutions.entrySet()) {
             LOGGER.log(Level.INFO, String.format("urlSubstitution: '%s' -> '%s'", entry.getKey(), entry.getValue()));
@@ -104,7 +109,15 @@ public class Ncwms {
         try {
             @SuppressWarnings("unchecked")
             Map<String, String[]> wmsParameters = new HashMap<String, String[]>(request.getParameterMap());
-            wmsParameters.remove("TIME");
+
+            // Some collections such as SRS are indexed with a timestamp which doesn't match
+            // the timestamp thredds calculates but only have one timestamp per file meaning its 
+            // not actually required
+            // For the moment just don't include the time parameter for these collections
+            if (isCollectionWithTimeMismatch(layerDescriptor.geoserverName())) {
+                wmsParameters.remove("TIME");
+            }
+
             wmsParameters.put("VERSION", new String[] { wmsVersion });
             wmsParameters.put(layerParameter, new String[] { layerDescriptor.variable });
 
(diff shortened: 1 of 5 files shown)


jonescc:
DON'T MERGE - looking at some issues with CARS World Monthly time step selection

jonescc:
Added an option to bypass forwarding the requested date and time onto thredds for collections which don't need to do so and where doing so would be problematic.

SRS has been indexed with date time values which don't match what thredds calculates meaning thredds rejects the date time requested as being invalid.  The date time is not required to be passed on for SRS anyway as each SRS file only contains data for a single date time.

Will look at whether we should reharvest SRS with the correct date time information separately.

jonescc:
OK to review/merge now.

julian1:
I'll review

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/Ncwms.java
```diff
@@ -104,7 +109,15 @@ private void proxyWmsRequest(HttpServletRequest request, HttpServletResponse res
         try {
             @SuppressWarnings("unchecked")
             Map<String, String[]> wmsParameters = new HashMap<String, String[]>(request.getParameterMap());
-            wmsParameters.remove("TIME");
+
+            // Some collections such as SRS are indexed with a timestamp which doesn't match
```
julian1: Should the harvester be changed to record instantaneous dates, as well as intervals when that's appropriate? Assuming I understand the issue.

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/NcwmsConfig.java
```diff
@@ -68,4 +69,27 @@ public String getConfigVariable(String xpathString, String defaultValue) {
...
+    public List<String> getConfigList(String xpathString) {
+        SAXReader reader = new SAXReader();
+
+        List<String> returnValue = new ArrayList<>();
+
+        try {
+            Document doc = reader.read(getConfigFile());
+            DefaultXPath xpath = new DefaultXPath(xpathString);
+
+            @SuppressWarnings("unchecked")
```
julian1: Why not catch and handle the required exceptions?

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/Ncwms.java
```diff
@@ -31,16 +33,19 @@
          <wfsServer>http://localhost:8080/geoserver/ows</wfsServer>
          <urlSubstitution key="/mnt/imos-t3/IMOS/opendap/">http://thredds-1-aws-syd.aodn.org.au/thredds/wms/IMOS/</urlSubstitution>
          <urlSubstitution key="^/IMOS/">http://thredds-1-aws-syd.aodn.org.au/thredds/wms/IMOS/</urlSubstitution>
+         <collectionsWithTimeMismatch>^imos:srs.*</collectionsWithTimeMismatch>
```
julian1: Would it better to call this disregardTimeField or something? to indicate what the effect of setting the flag is?

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/Ncwms.java
```diff
@@ -125,6 +138,16 @@ private void proxyWmsRequest(HttpServletRequest request, HttpServletResponse res
         }
     }
 
+    private boolean isCollectionWithTimeMismatch(String layerName) {
```
julian1: Are we java 1.8 now? 

Is it possible to write this as a filter and predicate? 

http://stackoverflow.com/questions/122105/what-is-the-best-way-to-filter-a-java-collection

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/NcwmsConfig.java
```diff
@@ -68,4 +69,27 @@ public String getConfigVariable(String xpathString, String defaultValue) {
...
+            List<DefaultElement> list = xpath.selectNodes(doc);
+
+            for (final DefaultElement element : list) {
+                returnValue.add(element.getText());
+            }
+        }
+        catch (DocumentException e) {
+            LOGGER.log(Level.WARNING, String.format("Could not open config file '%s': '%s'", getConfigFile(), e.getMessage()));
```
julian1: Is it possible to have DocumentException's that are something other than failing to open the config file?

julian1:
It's good that the behavior has been pushed out into config, but it still seems oddly specific to the data collection.

julian1:
No unit/ integration tests?

julian1:
finished!

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/NcwmsConfig.java
jonescc: Probably - I just copied Dan's code - I'll have a better look at this

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/NcwmsConfig.java
jonescc: Same here

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/Ncwms.java
jonescc: No not java 1.8 yet.

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/Ncwms.java
jonescc: Looking at changing the harvester - this is an interim fix as it will take quite some time to reharvest all the SRS data

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/UrlIndexFeatureSource.java
```diff
@@ -24,7 +24,6 @@
 import org.joda.time.DateTimeZone;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
-import org.joda.time.format.ISODateTimeFormat;
```
julian1: So this wasn't being used?

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/Ncwms.java
jonescc: I leant in favour of making it clear why the timefield was being disregarded rather than just saying to disregard it - I'd prefer to remove this code altogether and am pushing to reharvest all the data with the correct timestamps

src/extension/ncwms/src/main/java/au/org/emii/geoserver/wms/UrlIndexFeatureSource.java
jonescc: Lean't towards not removing this because of the comments it might draw, but did it anyway - its not used after all

jonescc:
OK have made some changes.  Adding integration tests would require a substantial amount of setup so perhaps that should be tackled as a backlog item.

jonescc:
@julian1 for action

julian1:
> @julian1 for action

rather bold.

PR #140, opened 2016-06-27
Subset multi-time files on requested time (GoGoDuck)

jonescc:
Pass time onto subset operations by default - required for multi time files such as CARS

Can be overridden on subclasses - i.e. SRS

Also changed location of CARS files to current location.  Its not ideal to do that here and should be changed when the location can be sourced form the database (required for ncwms abomination as well).


jonescc:
Needs some more testing

PR #143, opened 2016-07-04
Correct location of CARS files and time variable

jonescc:
Fixes subsetting on time for CARS


jonescc:
There's also an issue with the precision of the time subset passed by the portal which I will fix there.  Results in missing end times sometimes.

jonescc:
Ended up fixing the precision issue here as I imagine we would reduce the precision here to milliseconds if we made this a pure java solution.

jonescc:
Ready for review.

jonescc:
Rebased with master for testing purposes

PR #144, opened 2016-07-07
Fixes #2234

jonescc:
Don't request selected SRS variables for SRS_OC - it doesn't apply.

Fix required for master


jonescc:
Fixes aodn/aodn-portal#2234 as per fix made for systest https://github.com/aodn/geoserver-build/commit/732dceb5c65212267fbaa129730aedcd746ccd3d

PR #152, opened 2016-08-01
Multi time slice temp

bhasin85:


bhasin85:
@jonescc  I have created a new branch for the work I was doing.

src/extension/core/src/main/java/au/org/emii/core/Config.java
```diff
@@ -0,0 +1,163 @@
...
+    private GeoServerResourceLoader resourceLoader = null;
+    protected File resourceDirectory;
+    protected Catalog catalog;
+
+    public Config(File resourceDirectory, Catalog catalog) {
+        this.resourceDirectory = resourceDirectory;
+        this.catalog = catalog;
+    }
+
+    protected File getResourcesDirectory() {
```
jonescc: Not used

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/FeatureSourceIndexReader.java
```diff
@@ -80,4 +81,8 @@ private SimpleFeatureIterator getFeatures(String typeName, Query query) throws I
         SimpleFeatureCollection collection = featureSource.getFeatures(query);
         return collection.features();
     }
+
+    public Catalog getCatalog() {
```
jonescc: Not used?

bhasin85: Removed.

src/extension/core/src/main/java/au/org/emii/core/Config.java
bhasin85: Removed.

bhasin85:
@jonescc  Removed the time subset config.

src/extension/core/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+        <version>1.0.0</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <dependencies>
+        <!-- https://mvnrepository.com/artifact/dom4j/dom4j -->
+        <dependency>
+            <groupId>dom4j</groupId>
+            <artifactId>dom4j</artifactId>
+            <version>1.6.1</version>
```
jonescc: Should we use the geoserver managed version to avoid version issues?

src/extension/wps/doc/GOGODUCK_README.md
```diff
@@ -14,9 +14,11 @@ To build:
 $ mvn clean install
 ```
 
-To run:
+To build:
```
jonescc: I think "To run" made more sense here than "To build"

src/extension/wps/doc/GOGODUCK_README.md
```diff
@@ -14,9 +14,11 @@ To build:
...
 ```
-$ mvn exec:java -Dexec.mainClass=au.org.emii.gogoduck.worker.Main
+1. Navigate to http://localhost:9091/web/?wicket:bookmarkablePage=:org.geoserver.wps.web.WPSRequestBuilder
```
jonescc: 9091 may not be what other people have this running on.

jonescc: Assumes they have run mvn jetty:run - perhaps link to the section from here

src/extension/wps/doc/GOGODUCK_README.md
```diff
@@ -14,9 +14,11 @@ To build:
...
+2. Choose process: gs:GoGoDuck
+3. Choose format: application/x-netcdf
 ```
 
 ### Running as a WPS service. (simulating a portal)
```
jonescc: Perhaps "Submitting a request from the command line" would be a better heading - with "Submitting a request from Geoserver" being a better heading than "To build" above.

src/extension/wps/doc/GOGODUCK_README.md
```diff
@@ -32,43 +34,44 @@ $ mvn exec:java -Dexec.mainClass=au.org.emii.gogoduck.worker.Main
 
 ## Direct calls (using IMOS Data)
```
jonescc: I think these belong with the GeoServer option above now?

src/extension/core/src/main/java/au/org/emii/core/Config.java
```diff
@@ -0,0 +1,153 @@
+package au.org.emii.core;
```
jonescc: If this is just the ncwms abomination config class then perhaps we should remove the other and reference this one rather than just duplicating the code.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/Converter.java
```diff
@@ -2,6 +2,7 @@
 
 import java.nio.file.Path;
 
+import au.org.emii.gogoduck.exception.GoGoDuckException;
```
jonescc: This doesn't appear to be used.

jonescc:
src/extension/wps/output.csv and src/extension/wps/output.nc have been added but don't seem to be used anywhere?   Assuming these have been added accidently?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -39,25 +45,18 @@
     private ProgressListener progressListener = null;
     private String mimeType = "application/x-netcdf";
     private String extension = "nc";
+    private GoGoDuckConfig goGoDuckConfig;
 
-    protected GoGoDuck(String profile, String subset, String outputFile, String format, Integer limit) {
+    public GoGoDuck(Catalog catalog, String profile, String subset, String outputFile, String format, Integer limit, GoGoDuckConfig goGoDuckConfig) {
```
jonescc: Passing the file limit here twice?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -169,6 +172,8 @@ private void createSymbolicLink(File srcFile, Path dst) {
     }
 
     private void downloadFile(URI uri, Path dst) {
+        urlMangling.put(goGoDuckConfig.getUrlSubstitutionKey(), goGoDuckConfig.getUrlSubstitutionValue());
+        URLMangler.setUrlManglingMap(urlMangling);
         URL url = URLMangler.mangle(uri);
```
jonescc: Perhaps this should be instantiated with the urlmap rather than calling setting static maps here

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -169,6 +172,8 @@ private void createSymbolicLink(File srcFile, Path dst) {
     }
 
     private void downloadFile(URI uri, Path dst) {
+        urlMangling.put(goGoDuckConfig.getUrlSubstitutionKey(), goGoDuckConfig.getUrlSubstitutionValue());
```
jonescc: Can't there be more than one substituion in the map

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -234,18 +239,21 @@ private static void gunzipInPlace(File file) {
...
         try {
             File tmpFile = File.createTempFile("tmp", ".nc");
 
             List<String> command = new ArrayList<String>();
-            command.add(GoGoDuckConfig.ncksPath);
+            command.add(goGoDuckConfig.getNcksPath());
             command.add("-a");
             command.add("-4");
             command.add("-O");
+            if (!module.isTimeUnlimited()) {
```
jonescc: Doesn't read very well here - perhaps consider renaming module - what does it hold now -FileAggregationmetadata?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
jonescc: Perhaps we can also remove the setThreadLimit method too

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
jonescc: Actually, perhaps this should just be moved into the code in module that returns nckssubsetparameters.  Kind of confusing creating working out the subset parameters in two places.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -274,13 +282,15 @@ private static void applySubsetSingleFileNative(File file, GoGoDuckModule module
         private File file;
         private GoGoDuckModule module = null;
         private GoGoDuck gogoduck = null;
+        private List<String> ncksSubsetParameters = null;
         private ProgressListener progressListener = null;
 
-        NcksRunnable(File file, GoGoDuckModule module, GoGoDuck gogoduck) {
+        NcksRunnable(File file, GoGoDuckModule module, GoGoDuck gogoduck, List<String> ncksSubsetParameters) {
```
jonescc: Think this would cleaner if we just passed the module as noted above.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -291,20 +301,24 @@ public void run() {
...
 
         File[] directoryListing = tmpDir.toFile().listFiles();
-        logger.info(String.format("Subset for operation is '%s'", module.getNcksSubsetParameters()));
 
         ExecutorService executorService = Executors.newFixedThreadPool(threadCount);
+        List<String> ncksSubsetParameters = null;
         for (final File file : directoryListing) {
-            executorService.submit(new NcksRunnable(file, module, this));
+            if (ncksSubsetParameters == null) {
+                ncksSubsetParameters = module.getNcksSubsetParameters(file.getAbsoluteFile().toString()).getNcksParameters();
```
jonescc: Why not probe files for collection metadata in one place instead of two i.e. setup and here

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -316,31 +330,35 @@ private void applySubsetMultiThread(Path tmpDir, GoGoDuckModule module, int thre
...
         logger.info(String.format("Applying subset on directory '%s'", tmpDir));
 
         File[] directoryListing = tmpDir.toFile().listFiles();
-        logger.info(String.format("Subset for operation is '%s'", module.getNcksSubsetParameters()));
+        List<String> ncksSubsetParameters = null;
 
         for (File file : directoryListing) {
-            applySubsetSingleFileNcks(file, module);
+            if (ncksSubsetParameters == null){
+                ncksSubsetParameters = module.getNcksSubsetParameters(file.getAbsoluteFile().toString()).getNcksParameters();
```
jonescc: Do we need to probe the file again here?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -18,30 +32,107 @@
...
+                    timeUnlimited = false;
+                }
+
+                if (postProcessingRequired && !timeUnlimited) {
+                    break;
+                }
+            }
+            this.postProcessingRequired = postProcessingRequired;
+            this.timeUnlimited = timeUnlimited;
+            gridDs.close();
```
jonescc: This may never happen if there is an exception - needs to go into a finally block or a try resources block to appropriate cleanup.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -18,30 +32,107 @@
...
 
+    public boolean postProcessingRequired() {
+        return postProcessingRequired.booleanValue();
+    }
+
+    public boolean isTimeUnlimited() {
+        return timeUnlimited.booleanValue();
+    }
+
+    public void setup(File file) {
```
jonescc: Setup could do anything - perhaps renaming this would make it clearer for reading usage elsewhere - e.g. loadFileMetadata or something like that

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -18,30 +32,107 @@
...
+        boolean postProcessingRequired = false;
```
jonescc: Would renaming this to containsPackedVariable or something similar make it clearer why we are performing post processing?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -18,30 +32,107 @@
...
+
+    public void init(String profile, IndexReader indexReader, String subset, UserLog userLog, GoGoDuckConfig goGoDuckConfig) {
         this.profile = profile;
         this.indexReader = indexReader;
         this.subset = new GoGoDuckSubsetParameters(subset);
         this.userLog = userLog;
+        this.goGoDuckConfig = goGoDuckConfig;
     }
 
     public URIList getUriList() throws GoGoDuckException {
```
jonescc: Not sure this belongs here anymore - it was here before because of  hardcoded url's for CAR, but we are now going to the index for all files now.   Perhaps keep this class for sourcing collection specific aggegation metadata/config only and just use the indexreader directly in gogoduck.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -18,30 +32,107 @@
...
         return indexReader.getUriList(profile, TIME_FIELD, URL_FIELD, subset);
     }
 
+    public boolean postProcessingRequired() {
+        return postProcessingRequired.booleanValue();
```
jonescc: perhaps here test for the conditions to perform postprocessing i.e. hasPackedVariables

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -18,30 +32,107 @@
...
+            this.postProcessingRequired = postProcessingRequired;
+            this.timeUnlimited = timeUnlimited;
+            gridDs.close();
+        } catch (IOException e) {
+            logger.error(e.getMessage());
+            throw new GoGoDuckException(e.getMessage());
+        }
+    }
+
     public void postProcess(File file) {
```
jonescc: Perhaps move this to the gogoduck class to keep all actual file processing performed there and this class just for recording collection aggregation metadata

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -80,61 +171,61 @@ public void postProcess(File file) {
 
     public final void updateMetadata(Path outputFile) {
         try {
-            NetcdfFileWriter nc = NetcdfFileWriter.openExisting(outputFile.toAbsolutePath().toString());
+            String location = outputFile.toAbsolutePath().toString();
+            NetcdfFileWriter nc = NetcdfFileWriter.openExisting(location);
+            GridDataset gridDs = GridDataset.open(location);
 
             nc.setRedefineMode(true);
             for (Attribute newAttr : getGlobalAttributesToUpdate(nc.getNetcdfFile())) {
```
jonescc: Need to cater for collection specific metadata update such as that currently performed for SRS?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -18,30 +32,107 @@
...
+            GoGoDuck.execute(command);
+
+            Files.delete(file.toPath());
+            Files.move(tmpFile.toPath(), file.toPath());
+        } catch (Exception e) {
+            throw new GoGoDuckException(String.format("Could not run ncpdq on file '%s'", file.toPath()));
+        }
     }
 
     public List<String> ncksExtraParameters() {
```
jonescc: This was only used when time dimensions are not unlimited which we are handling elsewhere now - this option can be removed now?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -80,61 +171,61 @@ public void postProcess(File file) {
...
-            try {
-                Class classz = Class.forName(String.format("%s.%s", thisPackage, classToInstantiate));
-                module = (GoGoDuckModule) classz.newInstance();
-                module.init(profile, indexReader, subset, userLog);
-                logger.info(String.format("Using class '%s.%s'", thisPackage, classToInstantiate));
-                return module;
-            }
-            catch (Exception e) {
-                logger.debug(String.format("Could not find class for '%s.%s'", thisPackage, classToInstantiate));
+        try {
```
jonescc: Perhaps do this once in setup - we've got real problems if variables are named differently in each file.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule_srs.java
```diff
@@ -1,90 +0,0 @@
...
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import ucar.nc2.Attribute;
-import ucar.nc2.NetcdfFile;
-
-public class GoGoDuckModule_srs extends GoGoDuckModule {
-    private static final Logger logger = LoggerFactory.getLogger(GoGoDuckModule_srs.class);
-
-    private static final String srsVariables = "time,lat,lon,dt_analysis,l2p_flags,quality_level,satellite_zenith_angle,sea_surface_temperature,sses_bias,sses_count,sses_standard_deviation,sst_dtime";
```
jonescc: OK, so this is why you still  include extra parameters above, but I'd prefer to remove references to 'ncks parameters' from the config and replace them with with config that is more general such as variables to include.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule_srs.java
```diff
@@ -1,90 +0,0 @@
...
-                            title,
-                            subset.get("TIME").start,
-                            subset.get("TIME").end)));
-        }
-        catch (Exception e) {
-            // Don't fail because of this bullshit :)
-            logger.warn("Could not find 'title' attribute in result file");
-        }
-
-        newAttributeList.add(new Attribute("southernmost_latitude", subset.get("LATITUDE").start));
```
jonescc: I don't think we are handling this now

src/extension/wps/src/test/java/au/org/emii/gogoduck/worker/GoGoDuckModuleTest.java
```diff
@@ -1,37 +1,32 @@
...
-
-    public class GoGoDuckModule_test extends GoGoDuckModule {
-        @Override
-        protected List<Attribute> getGlobalAttributesToUpdate(NetcdfFile nc) {
-            List<Attribute> newAttributeList = new ArrayList<Attribute>();
-            newAttributeList.add(new Attribute("test_attribute", "test_value"));
-            return newAttributeList;
-        }
+        goGoDuckConfig = mock(GoGoDuckConfig.class);
+        ggdm = new GoGoDuckModule("", new HttpIndexReader(null, ""), "TIME,1,2;LONGITUDE,2,3", null, goGoDuckConfig);
```
jonescc: HttpIndexreader was required for the command line only I believe - perhaps we can remove it now

src/main/src/jetty/geoserver_data/workspaces/imos/geoserver/srs_timeseries_url/gogoduck.xml
```diff
@@ -0,0 +1,3 @@
+<gogoduck>
+    <ncks.parameters>-v;time,lat,lon,dt_analysis,l2p_flags,quality_level,satellite_zenith_angle,sea_surface_temperature,sses_bias,sses_count,sses_standard_deviation,sst_dtime</ncks.parameters>
```
jonescc: Lets not refer to ncks here - perhaps variables.

jonescc: I thought we were going to put this at the data store level - thats were it differs for all our datasets.

jonescc: Also perhaps break out into a separate element for each variable - comma separated list is ncks specific

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
jonescc: Actually - no still needed for specifying variables - but perhaps this should just be incorporated in a single getNcksSubsetParameters method.

jonescc:
Finished review of additional changes.

jonescc:
Had a look at TPAC data and looks like we should be able to automatically detect coordinate variables fine.  However, looks like its OK for TPAC files to remain packed as all files are packed in the same way (same scale_factor/offset) - so I think unpacking should be a config option for SRS at the datastore level rather than trying to automatically work this out (we would have to check the scale_factor/offset for all variables in all files before deciding whether to pack/unpack otherwise).  We would leave files packed by default.

src/extension/core/pom.xml
bhasin85: Make Sense.

src/extension/wps/doc/GOGODUCK_README.md
bhasin85: Updated

src/extension/core/src/main/java/au/org/emii/core/Config.java
bhasin85: I was planning to just update the wps extension and not touch ncwms for this task as it is out of  scope. Also, probably I will need to setup test data/environment for ncwms before updating it.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/Converter.java
bhasin85: Check line 29

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
bhasin85: Removed.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
bhasin85: Fixed.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
bhasin85: Moved following inside setup()

```
            this.postProcessingRequired = postProcessingRequired;
            this.timeUnlimited = timeUnlimited;
            setNcksSubsetParameters(file.getAbsoluteFile().toString());
            setNcksExtraParameters();
```

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
bhasin85: Fixed

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
bhasin85: Fixed

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
bhasin85: Added finally.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
bhasin85: Updated

src/extension/core/src/main/java/au/org/emii/core/Config.java
jonescc: I remember talking about this, although I was thinking more about not refactoring the netcdfgenerator wps to use the same config code as the config there is completely different.  As a general rule duplicating code is horrible - especially a straight copy and paste.  Not sure when we will ever refactor ncwms to use it if not now.   Perhaps something to look at when we finish the gogoduck refactor.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/Converter.java
jonescc: Yep your right.  How did this work before without the import?

bhasin85: Because GoGoDuckException was not getting used earlier.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
bhasin85: Updated.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
bhasin85: IndexReader method getUriList needs a parameter subset which is in GoGoDuckModule. To use  IndexReader inside GoGoDuck we will have to make indexReader and subset public. Also, I think it is not very friendly to call GoGoDuckModule variable method (IndexReader.getUriList) and passing GoGoDuckModule variable (subset) from GoGoDuck.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
bhasin85: Updated

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
bhasin85: Moved

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
jonescc: GoGoDuckModule doesn't have to own the GoGoDuckSubset object it can be instantiated in GoGoduck and passed to both the GoGoDuckModule and the UrlIndexReader.  - they are really doing separate things and they don't need to be in the same class anymore.  Having said that, its just cleaning up the code (the way I think it should be cleaned up) and is optional if you want to focus on getting the core processing to work again.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule_srs.java
bhasin85: Updated the name to variablesToInclude from ncks

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule_srs.java
bhasin85: Moved to config.

src/extension/wps/src/test/java/au/org/emii/gogoduck/worker/GoGoDuckModuleTest.java
bhasin85: Done

src/main/src/jetty/geoserver_data/workspaces/imos/geoserver/srs_timeseries_url/gogoduck.xml
bhasin85: Done.

bhasin85:
Checking for packedVariable config before working it out using API.

jonescc:
Given TPAC have been using default processing and they have been producing corrupted netcdf files (because they didn't unpack), I think unpacking should be the default if its packed unless specified otherwise.

src/extension/core/src/main/java/au/org/emii/core/Config.java
```diff
@@ -0,0 +1,154 @@
...
+        File storeDirectory = dataDirectory.findStoreDir(layerInfo.getResource().getStore());
+        String storeConfigFilePath = storeDirectory.toString() + "/" + configFileName;
+        if (new File(storeConfigFilePath).exists()) {
+            // returning relative path
+            return storeConfigFilePath.replace(this.resourceDirectory.getAbsolutePath().toString()+"/","");
+        }
+
+        // If config file not present in layer or store directory throw exception
+        throw new Exception("Unable to find config file for layer " + layer);
+    }
```
jonescc: Nice!

src/extension/wps/src/main/java/au/org/emii/wps/GoGoDuckProcess.java
```diff
@@ -63,87 +63,28 @@ public FileRawData execute(
...
                     getResourceManager().getExecutionId(true), layer + ".nc").file();
 
             String filePath = outputFile.toPath().toAbsolutePath().toString();
 
-            GoGoDuck ggd = new GoGoDuck(catalog, layer, subset, filePath, format, fileLimit);
+            GoGoDuck ggd = new GoGoDuck(catalog, layer, subset, filePath, format, goGoDuckConfig);
 
             ggd.setTmpDir(getWorkingDir());
-            ggd.setThreadCount(threadCount);
+            ggd.setThreadCount(goGoDuckConfig.getThreadCount());
```
jonescc: But we've already passed gogoduckconfig to gogoduck so we don't need to set the threadcount separately as well..

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -39,25 +44,22 @@
...
-        this(profile, subset, outputFile, format, limit);
-        this.indexReader = new HttpIndexReader(this.userLog, geoserverUrl);
-    }
-
-    public GoGoDuck(Catalog catalog, String profile, String subset, String outputFile, String format, Integer limit) {
-        this(profile, subset, outputFile, format, limit);
         this.indexReader = new FeatureSourceIndexReader(this.userLog, catalog);
+        this.goGoDuckConfig = goGoDuckConfig;
+        Map<String, String> urlMangling = new HashMap<>();
+        urlMangling.put(goGoDuckConfig.getUrlSubstitutionKey(), goGoDuckConfig.getUrlSubstitutionValue());
```
jonescc: There can be more than one

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -234,25 +228,29 @@ private static void gunzipInPlace(File file) {
...
+            }
+
+            command.addAll(module.getNcksSubsetParameters().getNcksParameters());
+            command.addAll(module.getNcksExtraParameters());
 
             command.add(file.getPath());
             command.add(tmpFile.getPath());
 
-            logger.info(String.format("Applying subset '%s' to '%s'", ncksSubsetParameters, file.toPath()));
+            logger.info(String.format("Applying subset '%s' to '%s'", module.getNcksExtraParameters(), file.toPath()));
```
jonescc: Shouldn't this be ncksSubsetParameters?

jonescc: Acttually no I can see you are logging it further down

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -291,100 +282,121 @@ public void run() {
...
-
-    private static void postProcess(Path tmpDir, GoGoDuckModule module) throws GoGoDuckException {
+    private void postProcess(Path tmpDir, GoGoDuckModule module) throws GoGoDuckException {
         File[] directoryListing = tmpDir.toFile().listFiles();
-        for (File file : directoryListing) {
-            module.postProcess(file);
+        if (directoryListing != null) {
+            for (File file : directoryListing) {
+                try {
+                    if (!module.hasPackedVariables()) {
```
jonescc: Why not do this at the start of the method - we know this before doing a directory listing

src/extension/wps/src/main/java/au/org/emii/wps/GoGoDuckProcess.java
bhasin85: Cleaned up.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
bhasin85: Updated the code

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
bhasin85: Moved the check to the beginning.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -1,52 +1,128 @@
...
     }
 
-    public List<String> ncksExtraParameters() {
-        List<String> ncksExtraParameters = new ArrayList<String>();
+    public void loadFileMetadata(File file) {
+        boolean containsPackedVariable = false;
+        GridDataset gridDs = null;
+
+        try {
+            setNcksSubsetParameters(file.getAbsoluteFile().toString());
```
jonescc: this is now poorly named - its doing two things 

a) probing the file to determine coordinate variables, and
b) constructing the subsetting parameters to be passed to ncks

this makes the reference to time.isUnlimited below confusing - I wonder where did time get set

jonescc: I'd split the two actions into a method to determine coordinate variables and a method to getnckssubsetparameters

getnckssubsetparameters can just be called where its needed

jonescc: Actually all its doing is probing for coordinate variables so a simple rename will do

jonescc: Deleting that last comment and this one - its doing two things

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -1,52 +1,128 @@
...
+            } catch (IOException e) {
+                logger.error(e.getMessage());
+                throw new GoGoDuckException(e.getMessage(), e);
+            }
+        }
+    }
+
+    private void setNcksExtraParameters() throws Exception {
+        ncksExtraParameters = new ArrayList<>();
+        for (String ncksParameter : goGoDuckConfig.getNcksParameters(profile)) {
```
jonescc: There's no need to have this in config anymore - if time is not unlimited we need to pass --mk_rec

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -66,75 +142,77 @@ public void postProcess(File file) {
                         subset.get("TIME").start,
                         subset.get("TIME").end)));
 
-        newAttributeList.add(new Attribute("geospatial_lat_min", subset.get("LATITUDE").start));
-        newAttributeList.add(new Attribute("geospatial_lat_max", subset.get("LATITUDE").end));
+        newAttributeList.add(new Attribute(goGoDuckConfig.getLatitudeStart(profile), subset.get("LATITUDE").start));
```
jonescc: The default should be geospatial_lat_min, geospatial_lat_max and so on. if not configured - lets not make them configure the default values

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -66,75 +142,77 @@ public void postProcess(File file) {
...
-        newAttributeList.add(new Attribute("geospatial_lon_max", subset.get("LONGITUDE").end));
+        newAttributeList.add(new Attribute(goGoDuckConfig.getLongitudeStart(profile), subset.get("LONGITUDE").start));
+        newAttributeList.add(new Attribute(goGoDuckConfig.getLongitudeEnd(profile), subset.get("LONGITUDE").end));
 
-        newAttributeList.add(new Attribute("time_coverage_start", subset.get("TIME").start));
-        newAttributeList.add(new Attribute("time_coverage_end", subset.get("TIME").end));
+        newAttributeList.add(new Attribute(goGoDuckConfig.getTimeStart(profile), subset.get("TIME").start));
+        newAttributeList.add(new Attribute(goGoDuckConfig.getTimeEnd(profile), subset.get("TIME").end));
 
         return newAttributeList;
```
jonescc: have we missed title here?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
bhasin85: There are other parameters needed to be passed like "time,lat,lon,dt_analysis,l2p_flags,quality_level,satellite_zenith_angle,sea_surface_temperature,sses_bias,sses_count,sses_standard_deviation,sst_dtime", therefore I believe we still need config.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
bhasin85: Seperated the code into separate functions.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
jonescc: OK yep forgot about them

src/extension/wps/src/main/java/au/org/emii/utils/GoGoDuckConfig.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public String getTimeField() {
+        return getConfig(TIME_FIELD, CONFIG_FILE);
+    }
+
+    public String getFileUrlField() {
+        return getConfig(FILE_URL_FIELD, CONFIG_FILE);
+    }
+
+    public List<String> getNcksParameters(String layer) throws Exception {
```
jonescc: This should be about getting the variable list

jonescc: Default if not specified is to retrieve all

src/extension/core/src/main/java/au/org/emii/core/Config.java
```diff
@@ -0,0 +1,154 @@
...
+        // Checking config file in store directory
+        File storeDirectory = dataDirectory.findStoreDir(layerInfo.getResource().getStore());
+        String storeConfigFilePath = storeDirectory.toString() + "/" + configFileName;
+        if (new File(storeConfigFilePath).exists()) {
+            // returning relative path
+            return storeConfigFilePath.replace(this.resourceDirectory.getAbsolutePath().toString()+"/","");
+        }
+
+        // If config file not present in layer or store directory throw exception
+        throw new Exception("Unable to find config file for layer " + layer);
```
jonescc: perhaps go to global config here

src/extension/wps/src/main/java/au/org/emii/utils/GoGoDuckConfig.java
```diff
@@ -0,0 +1,116 @@
...
+    private final String NCRCAT_PATH = FILE_PREFIX + "ncrcatPath";
+
+    private final String NCKS_PARAMETERS = FILE_PREFIX + "variablesToInclude";
+    private final String LATITUDE_START = FILE_PREFIX + "latitude.start";
+    private final String LATITUDE_END = FILE_PREFIX + "latitude.end";
+    private final String LONGITUDE_START = FILE_PREFIX + "longitude.start";
+    private final String LONGITUDE_END = FILE_PREFIX + "longitude.end";
+    private final String TIME_START = FILE_PREFIX + "time.start";
+    private final String TIME_END = FILE_PREFIX + "time.end";
+    private final String UNPACK_NETCDF = FILE_PREFIX + "unpack.netcdf";
```
jonescc: just 'unpack' would probably be better

src/extension/wps/src/main/java/au/org/emii/utils/GoGoDuckConfig.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private final String TIME_FIELD = FILE_PREFIX + "time.field";
```
jonescc: camelcase elementname using '.' is unusual

src/extension/wps/src/main/java/au/org/emii/utils/GoGoDuckConfig.java
```diff
@@ -0,0 +1,116 @@
...
+    private final String FILE_URL_FIELD = FILE_PREFIX + "file.url.field";
```
jonescc: and here

bhasin85:
@jonescc  I have made the discussed changes. Please review the changes. Thanks :smile:

src/main/src/jetty/geoserver_data/wps/gogoduck.xml
```diff
@@ -2,4 +2,19 @@
...
-</gogoduck>
\ No newline at end of file
+    <ncksPath>/usr/bin/ncks</ncksPath>
+    <ncpdqPath>/usr/bin/ncpdq</ncpdqPath>
+    <ncrcatPath>/usr/bin/ncrcat</ncrcatPath>
+    <timeField>time</timeField>
+    <fileUrlField>file_url</fileUrlField>
+    <unpack>true</unpack>
+    <globalAttributes>
+        <attribute key="title">title</attribute>
```
jonescc: no idea what an attribute key is - name is more readable for those who need to update this file.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -60,81 +126,83 @@ public void postProcess(File file) {
...
+                if (longitude == null) {
+                    longitude = grid.getCoordinateSystem().getXHorizAxis();
+                }
+
+                if (time != null && latitude != null && longitude != null) {
+                    break;
+                }
             }
-            classToInstantiate = nextProfile(classToInstantiate);
+            gridDs.close();
```
jonescc: should be in a finally clause or try resource block

src/main/src/jetty/geoserver_data/wps/gogoduck.xml
jonescc: OK, I see what you are doing now - I've been working at cross-purposes - lets discuss.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -1,52 +1,118 @@
...
+
+    public void loadFileMetadata(File file) {
+        boolean containsPackedVariable = false;
+        GridDataset gridDs = null;
+
+        try {
+            setTimeLatLon(file.getAbsoluteFile().toString());
+            if (goGoDuckConfig.getUnpackNetcdf(profile)) {
+                // Checking the config (Needed for TPAC)
+                this.containsPackedVariable = true;
```
jonescc: This will result in us trying to unpack files that have no packed variables and is a bit confusing because the file does not necessarily actually contain any packed variables.  Perhaps rename to unpack - and only set if the file contains packed variables and we have configured gogoduck to unpack packed variables.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuckModule.java
```diff
@@ -60,81 +126,83 @@ public void postProcess(File file) {
             logger.warn("Could not find 'title' attribute in result file");
         }
 
-        newAttributeList.add(new Attribute("title",
+        newAttributeList.add(new Attribute(goGoDuckConfig.getTitle(profile),
```
jonescc: Need to check what we can get away with here with @ggalibert

bhasin85:
@jonescc  Please review the changes. :+1:

PR #155, opened 2016-08-08
Fix for unpack config and extra logging

bhasin85:


src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -322,11 +322,14 @@ private void postProcess(Path tmpDir, GoGoDuckModule module) throws GoGoDuckExce
         if (directoryListing != null) {
             for (File file : directoryListing) {
                 try {
+                    logger.info(String.format("Checking file %s for packed variable", file.getAbsolutePath()));
                     if (!module.unpackNetcdf()) {
+                        logger.info(String.format("No packed variable found in %s", file.getAbsolutePath()));
```
jonescc: Perhaps 'Not unpacking {file name}... 'would be better - this gives the impression that there are no unpacked variables in the file - but this may not be the case - config could be overriding this.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -322,11 +322,14 @@ private void postProcess(Path tmpDir, GoGoDuckModule module) throws GoGoDuckExce
         if (directoryListing != null) {
             for (File file : directoryListing) {
                 try {
+                    logger.info(String.format("Checking file %s for packed variable", file.getAbsolutePath()));
```
jonescc: This is not really whats going on here - perhaps just delete - the two messages below should be sufficient

bhasin85: @jonescc  Please review the changes. Thanks!

PR #157, opened 2016-08-09
Fixes aodn/aodn-portal#2252

jonescc:
Process files in time order


src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -98,18 +98,14 @@ public Path run() {
...
-            downloadFiles(URIList, tmpDir);
+            List<Path> downloadedFiles = downloadFiles(URIList, tmpDir);
             throwIfCancelled();
-            File tempDirFile = new File(tmpDir.toString());
-            File[] files = tempDirFile.listFiles();
-            if (files != null && files.length > 0) {
-                module.loadFileMetadata(files[0]);
-            }
-            applySubsetMultiThread(tmpDir, module, goGoDuckConfig.getThreadCount());
+            module.loadFileMetadata(downloadedFiles.get(0).toFile());
```
bhasin85: Do you think it is a good idea to include the condition to check if at least one file is downloaded?

```
            if (files != null && files.length > 0) {
                 module.loadFileMetadata(files[0].toFile());
             }
```

jonescc: No, its done in enforce limits - we don't need to check it everywhere else

PR #165, opened 2016-08-31
fixes returning empty files

pmbohm:
Fixes https://github.com/aodn/aodn-portal/issues/2268

### DIFF
--- a/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
+++ b/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
@@ -5,6 +5,7 @@ import au.org.emii.gogoduck.exception.NetCdfProcessingException;
 import au.org.emii.utils.GoGoDuckConfig;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang.StringUtils;
 import org.geoserver.catalog.Catalog;
 import org.opengis.util.ProgressListener;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -409,25 +405,37 @@ public static int execute(List<String> command) throws Exception {
...
             String line;
+
             while ((line = br.readLine()) != null) {
                 logger.info(line);
+                bashLog.append(line);
             }
 
             process.waitFor();
 
+            if (process.exitValue() != 0) {
```
anguss00: I'm a bit concerned about the handling here of process. In my testing, the exception is thrown and bubbled up but by the time we have an exit value for the process, the damage has already been done and the file has been created (and therefore returned). So we get an exception in our catalina log. So, my suggestion would be to consider setting the processListener.isCancelled(true) whenever we throw a GGD exception. Needs testing and validation of course.

bhasin85: @anguss00  I am able to reproduce the issue using the portal. I am updating the fix.

anguss00: Ok, cool. Thanks @bhasin85

bhasin85: @anguss00  I tried replicating the [error](https://github.com/aodn/aodn-portal/issues/2263) with my changes without any success. I believe the exception handling will take care of it as earlier the errors were getting logged as info and some exceptions were suppressed. Let me know if you have any concern/question.

pmbohm: `processListener.isCancelled(true) ` worked well for me. I believed you were going to expand on this in other functions. As @anguss00 commented previously  `to consider setting the processListener.isCancelled(true) whenever we throw a GGD exception`

bhasin85: @pmbohm  @anguss00 

> To reproduce
> 
> Perform a subset where an ncks, npdqd or other command fails e.g. #2263

I think the changes fixes the issue.

PR #175, opened 2016-10-07
Updated to netcdf4 version 4.6.6

bhasin85:
Fix for error `ERROR [hdf5.H5header] - bad version 72 at filePos 218874346`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,13 +56,13 @@
             </dependency>
             <dependency>
                 <groupId>edu.ucar</groupId>
-                <artifactId>netcdf</artifactId>
-                <version>4.3.22</version>
+                <artifactId>netcdf4</artifactId>
+                <version>4.6.6</version>
             </dependency>
             <dependency>
                 <groupId>edu.ucar</groupId>
                 <artifactId>opendap</artifactId>
-                <version>4.3.22</version>
+                <version>4.6.6</version>
             </dependency>
             <dependency>
                 <groupId>commons-cli</groupId>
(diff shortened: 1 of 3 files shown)


src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/FileMetadata.java
```diff
@@ -5,9 +5,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import ucar.nc2.Attribute;
-import ucar.nc2.NetcdfFileWriter;
-import ucar.nc2.Variable;
+import ucar.nc2.*;
```
lwgordonimos: did you mean to do this?

pom.xml
```diff
@@ -56,13 +56,13 @@
...
-                <artifactId>netcdf</artifactId>
-                <version>4.3.22</version>
+                <artifactId>netcdf4</artifactId>
+                <version>4.6.6</version>
             </dependency>
             <dependency>
                 <groupId>edu.ucar</groupId>
                 <artifactId>opendap</artifactId>
-                <version>4.3.22</version>
+                <version>4.6.6</version>
```
lwgordonimos: Bear in mind that the servers are stuck on the 4.1.1 system library. Assuming these are just the Java bindings versions, do they still work as expected on our old libnetcdf6 version?

jonescc: Its not an issue here as we are not attempting to write or modify netcdf 4 files using java.  The netcdf java library only requires the netcdf c library to write to netcdf4 files and then there are no versions of the netcdf java library that could work with version 4.1.1 of the c library anyway (we know we tried at one point to update attributes and we had to revert to using nccatted instead).

If we wish to write to netcdf4 files using java we would need version 4.3.1 installed or greater.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/FileMetadata.java
jonescc: Updated

PR #177, opened 2016-10-12
Use s3 artifacts store example

jonescc:
Example of what may be required to use our geoserver fork containing queueing fixes and s3 artifact store.

Didn't get time to look at how we pull in other jenkins built artifacts so that will probably need to be adjusted.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
     <url>http://imos.org.au</url>
 
     <properties>
-        <geoserver.version>2.8.2</geoserver.version>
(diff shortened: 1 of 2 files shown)


jonescc:
Note that the temporaryStorageDir is the local storage directory for storing (for example) GoGoDuck temporary files and in this example is relative to the geoserver data directory - which is the default if omitted.   Can also be an absolute path.

jonescc:
Superceded by other pull requests

PR #182, opened 2016-10-19
Include s3 artifacts plugin

jonescc:


jonescc:
Requires https://github.com/aodn/chef-private/pull/2116 to build.

PR #184, opened 2016-10-21
Move Maven repo to content.aodn.org.au bucket

lwgordonimos:
Moved Maven repo to content.aodn.org.au for S3 consolidation.


lwgordonimos:
Have already copied repo content, so if Travis build is still happy, should be good to go.

PR #185, opened 2016-10-21
Improved logging and added informative error message for user

bhasin85:
Fixes [#2364](https://github.com/aodn/aodn-portal/issues/2364)


anguss00:
Too many files on the dance floor.
Too many files

PR #187, opened 2016-10-25
Added subset validation

bhasin85:
Fixes [#2366](https://github.com/aodn/aodn-portal/issues/2366)


src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -259,6 +262,12 @@ private void downloadFile(URL url, Path dst) {
             } catch (Exception e) {
                 logger.error(e.getMessage(), e);
             }
+
+            if (downloadedFiles.size() == 1) {
```
jonescc: So we'll only get metadata if we downloaded 1 file?

jonescc: Whoops - no read that wrong - apologies

PR #189, opened 2016-10-25
Update lon_max with correct value

jonescc:


bhasin85:
Thanks @jonescc  :+1:

PR #191, opened 2016-10-27
Added point time series validation

bhasin85:
Fixes [#2375](https://github.com/aodn/aodn-portal/issues/2375)


jonescc:
Looks good

PR #199, opened 2016-12-07
Allow more than 2232 files to be aggregated

jonescc:
Instead of passing the files to be aggregated on the command line (limited to 2232)
pass the files to be aggregated on stdin which has no such limit.

### DIFF
--- a/src/extension/wps/pom.xml
+++ b/src/extension/wps/pom.xml
@@ -95,6 +95,10 @@
             <artifactId>core</artifactId>
             <version>1.0.0</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-exec</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
--- a/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
+++ b/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
@@ -3,6 +3,11 @@ package au.org.emii.gogoduck.worker;
 import au.org.emii.gogoduck.exception.GoGoDuckException;
 import au.org.emii.gogoduck.exception.NetCdfProcessingException;
 import au.org.emii.utils.GoGoDuckConfig;
+import org.apache.commons.exec.CommandLine;
+import org.apache.commons.exec.DefaultExecutor;
+import org.apache.commons.exec.Executor;
+import org.apache.commons.exec.LogOutputStream;
+import org.apache.commons.exec.PumpStreamHandler;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang.StringUtils;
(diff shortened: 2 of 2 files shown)


src/extension/wps/pom.xml
```diff
@@ -95,6 +95,10 @@
             <artifactId>core</artifactId>
             <version>1.0.0</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-exec</artifactId>
```
julian1: does this want/need a version number?

jonescc: Its specified in geoserver - don't want to specify something different here

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -446,17 +450,20 @@ private void aggregate(List<Path> files, Path outputFile) throws GoGoDuckExcepti
...
+
+            String stdin = StringUtils.join(fileNames, " ");
 
             // Running ncrcat
             try {
-                execute(command);
+                execute(command, stdin);
             }
             catch (Exception e) {
                 logger.error(String.format("Could not concatenate files: %s into a single file: '%s'", Arrays.toString(fileNames.toArray()), outputFile));
```
julian1: This will lead to really large log entries. Is that an issue?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -497,6 +504,22 @@ protected void cleanTmpDir(Path tmpDir) {
...
+    private void execute(CommandLine command, String stdin) {
+        InputStream inputStream = new ByteArrayInputStream(stdin.getBytes());
+
+        CollectingLogOutputStream outputStream = new CollectingLogOutputStream();
+
+        try {
+            Executor executor = new DefaultExecutor();
+            executor.setStreamHandler(new PumpStreamHandler(outputStream, outputStream, inputStream));
+            executor.execute(command);
+        } catch (IOException ee) {
```
julian1: why ee rather than e? 

and should it be passed to NetCdfProcessingException when we rethrow NetCdfProcessingException

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -540,4 +563,18 @@ public static int execute(List<String> command) throws Exception {
     public String getMimeType() { return mimeType; }
 
     public String getExtension() { return extension; }
+
+    private static class CollectingLogOutputStream extends LogOutputStream {
```
julian1: What is the reason for making this helper class static?

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
jonescc: Used to be for execution exception I think - I didn't rename it.

Yes it should be.  I'll change it.

src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
jonescc: Yes it will - I'll remove this logging as the GoGoDuckException is logged elsewhere

jonescc:
@julian1  changes made

julian1:
Looks a lot cleaner!

PR #200, opened 2016-12-11
ncrcat writes everything to stderr so just log all output to info

jonescc:


src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java
```diff
@@ -505,19 +505,16 @@ protected void cleanTmpDir(Path tmpDir) {
...
-                    logger.info(line);
-                }
-            };
-            LogOutputStream errorStream = new LogOutputStream() {
-                @Override protected void processLine(String line, int level) {
-                    logger.error(line);
-                }
-            }
-        ){
+            LogOutputStream outputStream = new InfoLogOutputStream();
```
julian1: Is it possible to instantiate just one instance of InfoLogOutputStream, and then pass it to the setStreamHandler() for both output and error?

Unless there's an issue with concurrency? It doesn't look like the override of processLine is synchronized so I'm not sure it would make any difference?

PR #201, opened 2016-12-11
Make the most of the ncrcat stdin file list limit

jonescc:
Increases number of files which can be handled by gogoduck from 50000 up to over 150000 files more than enough for existing collections

Also reduces the amount of memory used (more could be done here).


julian1:
Looks like a cleaner way to handle the file passing as well. I'm surprised that ncrcat limits are not configurable.

PR #204, opened 2017-01-30
Add parallel download supporting classes

jonescc:


jonescc:
Need to fix failing test on travis

PR #205, opened 2017-01-30
Add parallel download supporting classes

jonescc:


src/extension/wps/src/main/java/au/org/emii/gogoduck/download/Config.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public static class ConfigBuilder {
+        private Path downloadDirectory;
+        private long localStorageLimit;
+        private int downloadAttempts;
+
+        public ConfigBuilder() {
+            downloadDirectory = Paths.get(System.getProperty("java.io.tmpdir"));
+            localStorageLimit = 100 * 1024 * 1024; // 100MiB
+            downloadAttempts = 3;
```
bhasin85: Do you think it will be better to move this to config and access it via au.org.emii.core.Config.

```
          localStorageLimit = 100 * 1024 * 1024; // 100MiB
           downloadAttempts = 3;
```

src/extension/wps/src/main/java/au/org/emii/gogoduck/download/Config.java
```diff
@@ -0,0 +1,62 @@
+package au.org.emii.gogoduck.download;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * Configuration setting for the ParallelDownloadManager
+ */
+public class Config {
```
bhasin85: We have a class with similar name au.org.emii.core.Config. Maybe we can rename it to something like DownloadConfig

src/extension/wps/src/main/java/au/org/emii/gogoduck/download/ParallelDownloadManager.java
```diff
@@ -0,0 +1,171 @@
...
+
+        private Download download() {
+            Path path = config.getDownloadDirectory().resolve(UUID.randomUUID().toString());
+
+            try {
+                downloader.download(request.getUrl(), path);
+                return new Download(request.getUrl(), path, request.getSize());
+            } catch (IOException e) {
+                logger.warn(String.format("Could not download %s", request.getUrl()), e);
+                return new DownloadError(request.getUrl(), path, request.getSize(), e);
```
bhasin85: So we will be continuing the processing even when some files fail to download. Which means same requests can give different results to the user. Am I getting it right?

src/extension/wps/src/test/java/au/org/emii/gogoduck/download/ParallelDownloadManagerTest.java
```diff
@@ -0,0 +1,235 @@
...
+    private Path downloadDir;
+    private Downloader downloader;
+    private ExecutorService pool;
+
+    @Before
+    public void setup() throws IOException {
+        downloadDir = Files.createTempDirectory("download.");
+        downloader = new Downloader(60*MILLISECONDS_IN_A_SECOND, 60*MILLISECONDS_IN_A_SECOND);
+        pool = Executors.newFixedThreadPool(8);
+    }
```
bhasin85: Maybe we can move this to config and access it via au.org.emii.core.Config. It is fine here for testing purposes.

src/extension/wps/src/main/java/au/org/emii/gogoduck/download/DownloadError.java
```diff
@@ -0,0 +1,21 @@
...
+import java.io.IOException;
+import java.net.URL;
+import java.nio.file.Path;
+
+/**
+ * Download error metadata
+ */
+public class DownloadError extends Download {
+    private final IOException exception;
+
```
bhasin85: How about introducing new exception type DownloadException and using it here? The code will still work though.

src/extension/wps/src/main/java/au/org/emii/gogoduck/download/Config.java
bhasin85: We can also get rid of Config and ConfigBuilder altogether and use au.org.emii.core.Config to read config to keep it consistant.

src/extension/wps/src/main/java/au/org/emii/gogoduck/download/ParallelDownloadManager.java
```diff
@@ -0,0 +1,171 @@
...
+            final DownloadRequest request = unactionedQueue.pop();
+            logger.debug(String.format("Adding %s to download queue", request.getUrl()));
+            inProgressQueue.add(pool.submit(new DownloadCallable(request)));
+            localStorageAllocated += request.getSize();
+            next = unactionedQueue.peek();
+        }
+    }
+
+    private boolean wontExceedLocalStorageLimit(DownloadRequest next) {
+        return localStorageAllocated + next.getSize() < config.getLocalStorageLimit();
```
jonescc: No problem - I'll update this

src/extension/wps/src/main/java/au/org/emii/gogoduck/download/ParallelDownloadManager.java
jonescc: It attempts to do the download up to the configured amount of times and then returns an error.  Its up to the client code what it does when an error happens - i.e gogoduck

src/extension/wps/src/main/java/au/org/emii/gogoduck/download/Config.java
jonescc: Sure - yes it is the downloadconfig

src/extension/wps/src/test/java/au/org/emii/gogoduck/download/ParallelDownloadManagerTest.java
jonescc: There will be a donloadconfigurer class when I hook this up in gogoduck injected by spring - wait tiull then to see how it all gets hooked up.

src/extension/wps/src/main/java/au/org/emii/gogoduck/download/DownloadError.java
jonescc: That sounds better - I'll have a look.

src/extension/wps/src/main/java/au/org/emii/gogoduck/download/Config.java
jonescc: Well - I was writing this code stand alone - when its hooked up to gogoduck this config will be populated from the gogoduck xml - will reconsider then.  The had coded values are defaults - mostly for testing purposes - will be populated from gogoduck config when its hooked up.

jonescc:
@bhasin - changes made - now throwing a DownloadException after retrying the configured number of times.

jonescc:
Also added a retryInterval configuration option for the amount of time to wait before retrying downloads.

PR #207, opened 2017-01-30
Enable logging in unit tests by removing extra slf4j provider

jonescc:


bhasin85:
@jonescc  Can you please also change the logging level to info as it will become so much easy to track the progress of the tests through command line.

jonescc:
It is info:

https://github.com/aodn/geoserver-build/blob/601a3caa47cff57053a04395c4f379447b293a51/src/extension/wps/src/test/resources/log4j.properties#L2

PR #208, opened 2017-02-08
Add variable unpacker for use in java aggregator

jonescc:


src/extension/wps/src/main/java/au/org/emii/aggregator/variable/UnpackerOverrides.java
```diff
@@ -0,0 +1,114 @@
...
+        private Number newValidMax;
+        private Number[] newMissingValues;
+        private DataType newDataType;
+
+        public Builder newFillerValue(Number newFillerValue) {
+            this.newFillerValue = newFillerValue;
+            return this;
+        }
+
+        public Builder newValidRange(Number[] newValidRange) {
```
julian1: If range is limited to two values, then min max variables might be better than using an array.

src/extension/wps/src/main/java/au/org/emii/aggregator/variable/UnpackerOverrides.java
```diff
@@ -0,0 +1,114 @@
...
+        this.newValidRange = builder.newValidRange == null ? null :
+            Arrays.copyOf(builder.newValidRange, builder.newValidRange.length);
+        this.newValidMin = builder.newValidMin;
+        this.newValidMax = builder.newValidMax;
+        this.newMissingValues = builder.newMissingValues == null ? null :
+            Arrays.copyOf(builder.newMissingValues, builder.newMissingValues.length);
+        this.newDataType = builder.newDataType;
+    }
+
+    public Number getNewFillerValue() {
```
julian1: There's a lot of type erasure here - using the Number superclass

- Are the encoding algorithms going to be allocating (eg. calling new) for every data value that has to be processed? GC will obviously handle it, but in terms of performance it might be better to pre-allocate once, and then reuse the objects.

- It's a pity Java's generic support isn't very strong, other wise you could parametize the encoders by netcdf data type (short, int, float, double), and handle things that way.

src/extension/wps/src/main/java/au/org/emii/aggregator/variable/datatype/FloatDataType.java
```diff
@@ -0,0 +1,23 @@
+package au.org.emii.aggregator.variable.datatype;
+
+/**
+ *Float DataType helper
+ */
+public class FloatDataType implements NumericType {
+    static private final float NC_FILL_FLOAT = 9.9692099683868690e+36f; /* near 15 * 2^119 */
```
julian1: Is this not already defined somewhere - in the netcdf libraries?

julian1: as a fill value convention?

jonescc: Its defined as private constants and duplicated in various places.

src/extension/wps/src/main/java/au/org/emii/aggregator/variable/UnpackerOverrides.java
jonescc: I benchmarked using primitive types/autoboxing and there was not enough of a difference to justify all the extra code I'd have to write (I couldn't see any difference actually)

src/extension/wps/src/main/java/au/org/emii/aggregator/variable/datatype/FloatDataType.java
jonescc: Actually just found a public defn - I'll use that

jonescc:
@julian1 changes made - I've stayed with using Number[] for the validrange in preference to creating a range class or using the commons range class for the moment.

PR #209, opened 2017-02-23
Add NetCDF Aggregator classes

jonescc:
OK so this works in so far as all the tests I've written work I checked them all manually.  Still plenty of scope for improvement but would be good to get some feedback via a review instead of me continually refining this.  

I'll finish off integrating this and parallel downloading into the gogoduck process.


akashisama:
ignore this,  testing to see if comments get added to my pending review for later publication or if they get published straight away

src/extension/wps/src/main/java/au/org/emii/aggregator/NetcdfAggregator.java
```diff
@@ -0,0 +1,331 @@
...
+                Array data = srcVariable.read(srcOrigin, slice);
+                writer.write(destVariable, destOrigin, data);
+                destOrigin[0]++;
+            }
+        } catch (IOException |InvalidRangeException e) {
+            throw new AggregationException(e);
+        }
+    }
+
+    private Map<String, UnpackerOverrides> getUnpackerOverrides() {
```
akashisama: By my reading, once `templateDataset` has been set up this method will return the same thing every time it's called no?
Could the overrides be placed in a variable and accessed from there rather than being worked out separately each time?

jonescc: Unpackeroverrides are really to do with reading data from the source files - I don't think they should be sourced from the template dataset now.  I've made some changes here in later pull requests which I'd like to build on.  So will tackle separately to this PR/after other PR's are reviewed/merged.

src/extension/wps/src/main/java/au/org/emii/aggregator/variable/NetcdfVariableAdapter.java
```diff
@@ -0,0 +1,61 @@
...
+import ucar.nc2.dataset.CoordinateAxis;
+import ucar.nc2.dataset.VariableDS;
+
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Convert a ucar.m2.VariableDS to the NetcdfVariable interface used in this library
+ */
+public class NetcdfVariableAdapter extends AbstractVariable {
```
jonescc: Need to think about this some more.  If the adapter is built into either the unpacker or subsetter then they will be forced to deal with concrete Unidata Variables making unit testing more difficult and requiring the use of a particular class.  I was trying to keep this generic and useful for other use cases.  Will handle any changes seperately to this PR.

PR #211, opened 2017-03-01
Allow configuration of new features

jonescc:
FOR REVIEW, but pull request is for merging into convert-gogoduck2 branch which may change while addressing review comments there so don't merge.


jonescc:
@akashisama - code has been rebased against javaduck-test - should be OK to merge when checks complete.

PR #216, opened 2017-03-14
Fix issue appending file multiple times when file contains more than …

jonescc:
…one time value.

Override equals/hasCode methods so linked hashset will not add the same download more than once.

### DIFF
--- a/src/extension/wps/src/test/java/au/org/emii/download/DownloadRequestTest.java
+++ b/src/extension/wps/src/test/java/au/org/emii/download/DownloadRequestTest.java
@@ -1,0 +1,32 @@
+package au.org.emii.download;
+
+import org.junit.Test;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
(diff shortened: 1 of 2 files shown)


src/extension/wps/src/test/java/au/org/emii/download/DownloadRequestTest.java
```diff
@@ -0,0 +1,31 @@
...
+        downloads.add(new DownloadRequest(new URL("http://example.com/downloads/file1.text"), 5325326L));
+
+        assertArrayEquals(new DownloadRequest[]{
+                new DownloadRequest(new URL("http://example.com/downloads/file1.text"), 5325326L),
+                new DownloadRequest(new URL("http://example.com/downloads/file2.text"), 532532L)
+            }, downloads.toArray()
+        );
+    }
+
+}
```
bhasin85: All good. Just put an extra line here. :+1:

jonescc: OK, done.

PR #221, opened 2017-04-11
Don't limit point subsets

jonescc:


src/extension/wps/src/main/java/au/org/emii/wps/GoGoDuckProcess.java
```diff
@@ -132,7 +130,12 @@ public FileRawData execute(
         }
     }
 
-    private void enforceFileLimits(Set<DownloadRequest> downloadList, Integer limit, double fileSizeLimit) throws GoGoDuckException {
+    private void enforceFileLimits(SubsetParameters parameters, Set<DownloadRequest> downloadList, Integer limit, double fileSizeLimit) throws GoGoDuckException {
```
bhasin85: Maybe we can just pass the boolean parameters.isPointSubset() instead of all the parameters.

jonescc: Well limits will be tested using all subset parameters in future.

PR #226, opened 2017-05-03
Change compression level

jonescc:
Change compression settings for faster aggregation time/smaller output file size as per outcome of [AWS Benchmarking](https://github.com/aodn/internal-discussions/issues/351).

To be included in next round of system testing!


src/extension/wps/src/main/java/au/org/emii/aggregator/NetcdfAggregator.java
```diff
@@ -120,8 +123,9 @@ private void assertOutputPathValid(Path outputPath) {
 
     private void copyToOutputFile(NetcdfDatasetIF template) throws AggregationException {
         try {
-            writer = NetcdfFileWriter.createNew(NetcdfFileWriter.Version.netcdf4,
-                outputPath.toString(), null);
+            Nc4Chunking chunking = new Nc4ChunkingDefault(3, false);
```
bhasin85: If the code is to be used by someone else, then it will be good if we make it configurable and move it to gogoduck.xml 


Does that make sense?

ggalibert:
Based on Craig's benchmarking and past experiences, I don't think anyone will want to change this anyway.

src/extension/wps/src/main/java/au/org/emii/aggregator/NetcdfAggregator.java
jonescc: Yes it does and I did consider it.  You could conceivably need/want to configure it at the variable, dataset and/or global level.  This would add a bit of complexity/work which would blow out the scope of this change.

Issue #229, opened 2017-05-09
Variable filler, min, max and valid range types should default to variable type if not specified


jonescc:
### To Reproduce

Specify an aggregation override like the following in an aggregation template:

```
        <variable name="sea_surface_temperature" type="Float">
            <attribute name="_FillValue" value="9.9692099683868690e+36"/>
            <attribute name="valid_min" value="0.0"/>
            <attribute name="valid_max" value="350.0"/>
        </variable>
```

### What happens

```
java.lang.NullPointerException
	at au.org.emii.aggregator.datatype.NumericTypes.get(NumericTypes.java:10)
	at au.org.emii.aggregator.datatype.NumericTypes.parse(NumericTypes.java:51)
	at au.org.emii.aggregator.overrides.VariableAttributeOverride.getAttributeNumericValue(VariableAttributeOverride.java:46)
	at au.org.emii.aggregator.overrides.VariableOverrides.getAttributeNumericValue(VariableOverrides.java:88)
	at au.org.emii.aggregator.overrides.VariableOverrides.getFillerValue(VariableOverrides.java:45)
	at au.org.emii.aggregator.NetcdfAggregator.getUnpackerOverrides(NetcdfAggregator.java:249)
```
### What I expect to happen

The type of the fill_value, valid_min, valid_max is set to the variable type

jonescc:
Fixed by  https://github.com/aodn/geoserver-build/commit/fe0b35bc01a7e6ff8aa016e8f6d854dae4d71b76

PR #233, opened 2017-05-30
Bug 229

bhasin85:
#229


src/extension/wps/src/main/java/au/org/emii/aggregator/overrides/VariableOverrides.java
```diff
@@ -80,12 +80,18 @@ public String toString() {
 
     private Number[] getAttributeNumericValues(String name) {
         VariableAttributeOverride attributeOverride = findAttribute(name);
-        return attributeOverride == null ? null : attributeOverride.getAttributeNumericValues(name);
+        return attributeOverride == null ? null : attributeOverride.getAttributeNumericValues();
```
jonescc: Probably need to cater for null type here as well.

src/extension/wps/src/test/java/au/org/emii/aggregator/NetcdfAggregatorTest.java
```diff
@@ -206,22 +206,32 @@ public void testLatLonProjectionSpatialSubsetNoData() throws IOException, Aggreg
         assertNetcdfFilesEqual(resourcePath("au/org/emii/aggregator/projection-point-outside-expected.nc"), outputFile);
     }
 
-    @Test
-    public void testAggregationOverrides() throws IOException, AggregationException {
+    private void testAggregationOverrides(String resourcePath) throws IOException, AggregationException {
```
jonescc: Minor thing - I'd normally put this below the methods that use it.  Normally you read java from top to bottom - higher level of abstraction at the top.

src/extension/wps/src/test/java/au/org/emii/aggregator/NetcdfAggregatorTest.java
```diff
@@ -206,22 +206,32 @@ public void testLatLonProjectionSpatialSubsetNoData() throws IOException, Aggreg
...
             netcdfAggregator.add(resourcePath("au/org/emii/aggregator/srs-2.nc"));
         }
 
         assertNetcdfFilesEqual(resourcePath("au/org/emii/aggregator/overrides-expected.nc"), outputFile);
+
+    }
+
+    @Test
+    public void testAggregationOverridesWithAttributeDataType() throws IOException, AggregationException {
+        testAggregationOverrides("au/org/emii/aggregator/template.xml");
```
jonescc: This will mean we'll now have to keep the two templates in sync  i.e. make the same change in two places.  I'd put the twoi different ways to specify the types on attributes in the one template.

src/extension/wps/src/main/java/au/org/emii/aggregator/overrides/VariableOverrides.java
bhasin85: Updated the check.

src/extension/wps/src/test/java/au/org/emii/aggregator/NetcdfAggregatorTest.java
bhasin85: Rearranged.

src/extension/wps/src/test/java/au/org/emii/aggregator/NetcdfAggregatorTest.java
bhasin85: I think keeping different template files will give us the flexibility to test different scenarios without making many changes to the tests.

bhasin85:
@jonescc Ready for review

src/extension/wps/src/test/resources/au/org/emii/aggregator/templateWithoutAttributeDataType.xml
```diff
@@ -0,0 +1,73 @@
...
+        </variable>
+
+        <variable name="sses_bias" type="Float">
+            <attribute name="valid_min">
+                <value>-10.0</value>
+            </attribute>
+            <attribute name="valid_max">
+                <value>10.0</value>
+            </attribute>
+            <attribute name="valid_range">
```
jonescc: Don't think this is actually being tested.  If it was it would have required a change to the expected dataset and to the other template to match (the problem with duplicating the template this way).

PR #235, opened 2017-05-31
Fixes #231

### LINKED ISSUE #231: JavaDuck on CARS without geographic subset
When downloading the full CARS monthly on step 3 (no temporal or geographical subset applied), the file produced only contains longitude values from 0 to 180 while the original file includes values from 0 to 360!

screenshot from 2017-05-22 17-18-06

jonescc:
Use netcdf bbox and latlonpointimpl for lat/lon comparisons as they handle mapping of longitude values to -180 to 180 range


jonescc:
refactoring this slightly to remove duplicated code

PR #237, opened 2017-06-02
fix #2480

julian1:


julian1:
Fixes the original complaint, https://github.com/aodn/aodn-portal/issues/2480. There is a remaining content problem that the query for anmn_timeseries performs too slowly and ends up using all the memory and killing geoserver.

julian1:
Closing, since it appears there are other null statements that get hit - after the sql queries are fixed for performance.

PR #239, opened 2017-06-08
Fix 2480

jonescc:
Handle null cql filter.  Use inner joins/filtering on instance.id instead of data.instance_id to improve performance.


akashisama:
Fixes https://github.com/aodn/aodn-portal/issues/2480

PR #243, opened 2017-06-29
Handle gridded data without time axis gracefully - return empty dates…

jonescc:
… rather than throwing a null pointer exception


jonescc:
Bump.  This one shouldn't be too onerous to review.

anguss00:
👍  pretty low risk ;-)

Issue #198, opened 2016-11-02
Improve GoGoDuck parameter handling


jonescc:
From the geoserver-wps logs:

```
02 Nov 00:51:20 INFO [worker.GoGoDuck] - Validating subset TIME,2014-10-10T00:00:00,2014-10-12T00:00:00;LATITUDE,-33.433849,-32.150743;LONGITUDE,114.15197,115.741219
02 Nov 00:51:20 INFO [worker.GoGoDuck] - Matched Time Pattern: 2014-10-10T00:00:00
02 Nov 00:51:20 INFO [worker.GoGoDuck] - Matched Time Pattern: 2014-10-12T00:00:00
02 Nov 00:51:20 INFO [worker.GoGoDuck] - Matched Latitude/Longitude Pattern: 00,2014
02 Nov 00:51:20 INFO [worker.GoGoDuck] - Matched Latitude/Longitude Pattern: -33.433849,-32.150743
02 Nov 00:51:20 INFO [worker.GoGoDuck] - Matched Latitude/Longitude Pattern: 114.15197,115.741219
02 Nov 00:51:20 ERROR [worker.GoGoDuck] - Your aggregation failed! Reason for failure is: 'Invalid latitude/longitude format for subset: TIME,2014-10-10T00:00:00,2014-10-12T00:00:00;LATITUDE,-33.433849,-32.150743;LONGITUDE,114.15197,115.741219'
```

Looks like GoGoduck relies on the time zone (e.g. Z) to be specified on dates to be able to correctly validate latitude/longitude (https://github.com/aodn/geoserver-build/blob/master/src/extension/wps/src/main/java/au/org/emii/gogoduck/worker/GoGoDuck.java#L148).  

Not sure how someone would work this out from the error returned.  Looks like the requestor spent a bit of time changing the number of decimals on the lat/longitude parameters to get their request to work without success.

jonescc:
Maybe https://github.com/aodn/geoserver-build/issues/150 would help

pblain:
@jonescc - is this still relevant for the Java version?

jonescc:
I haven't revisited parameter validation so still relevant.

pmbohm:
Fixed by https://github.com/aodn/geoserver-build/issues/245

PR #245, opened 2017-07-04
Updated regex for time validation

bhasin85:
#198

### DIFF
--- a/src/extension/wps/src/main/java/au/org/emii/wps/gogoduck/parameter/SubsetParameters.java
+++ b/src/extension/wps/src/main/java/au/org/emii/wps/gogoduck/parameter/SubsetParameters.java
@@ -45,6 +45,7 @@ public class SubsetParameters {
         Map<String, ParameterRange> subsets = new HashMap<>();
         Matcher matcher;
 
+
         // Parse
         for (String part : subset.split(";")) {
             String[] subsetParts = part.split(",");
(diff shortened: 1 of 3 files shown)


src/extension/wps/src/main/java/au/org/emii/wps/gogoduck/parameter/SubsetParameters.java
```diff
@@ -73,6 +64,7 @@ public static SubsetParameters parse(String subset) {
         ParameterRange timeRange = subsets.get("TIME");
         CalendarDateRange calendarDateRange = null;
 
+        //Validation
```
jonescc: Perhaps include expected date/time format in exception so requestor has some idea what they need to provide.

Could also include this in the process documentation included in https://github.com/aodn/geoserver-build/blob/master/src/extension/wps/src/main/java/au/org/emii/wps/GoGoDuckProcess.java#L55 (appears in WPS DescribeProcess request)

jonescc:
Looks good.  One minor comment.

src/extension/wps/src/main/java/au/org/emii/wps/gogoduck/parameter/SubsetParameters.java
bhasin85: @jonescc  Made the exception more descriptive. :+1:

PR #246, opened 2017-07-05
EnabledFilters Extension: Merge workspace and layer parameters

akashisama:
Allow workspace and layer name to be specified together in a single parameter (but still accept old way as well). Return a content-type so we're conforming to web standards and so geowebcache doesn't get whiney

### DIFF
--- a/src/extension/layer-filters/src/main/java/au/org/emii/geoserver/extensions/filters/LayerFiltersService.java
+++ b/src/extension/layer-filters/src/main/java/au/org/emii/geoserver/extensions/filters/LayerFiltersService.java
@@ -187,6 +187,7 @@ public class LayerFiltersService {
     }
 
     private void respondWithDocument(HttpServletResponse response, Document document) throws TransformerException, IOException {
+        response.addHeader("Content-Type", "text/xml");
         TransformerFactory transformerFactory = TransformerFactory.newInstance();
         Transformer transformer = transformerFactory.newTransformer();
         DOMSource source = new DOMSource(document);


src/extension/layer-filters/src/main/java/au/org/emii/geoserver/extensions/filters/LayerFiltersService.java
```diff
@@ -133,6 +133,11 @@ public void enabledFilters(HttpServletRequest request, HttpServletResponse respo
         String workspace = request.getParameter("workspace");
         String layer = request.getParameter("layer");
 
+        if(workspace==null) {
+            workspace = layer.split(":")[0];
```
lwgordonimos: What if there is a no 'workspace' parameter, and a 'layer' parameter with no colon in it? 

'workspace' is set to what is actually the layer name and the second split line throws an ArrayIndexOutOfBoundsException? Or am I misreading something?

akashisama: Yeah it will fall over, that's the preexisting behaviour. All the functions in the extension expect both a workspace and a layer name, so we'll have to rewrite them if we want to support layers without a workspace, which doesn't seem worth the effort.

lwgordonimos: Can it at least fall over at this point with a handled exception? ;)

Looks like in that file only 'enabledFilters' and 'uniqueValues' are actually dealing with URL parameters via the HttpServletRequest, so could it not have a common method of splitting the workspace from the URL and failing cleanly if a) there is a missing workspace parameter AND b) the layer name doesn't contain the workspace, and then sending a coherent HTTP response instead of just remaining unhandled? Doesn't seem like too much effort if it's only in that one file!

akashisama:
Actually, after looking into making the changes in response to @lwgordonimos's review I discovered that the extension is _already_ dealing with workspace being null, at a later stage. So only the content-type change is needed. I've altered the commit accordingly.

PR #247, opened 2017-07-05
Pass thredds version specified in the metadata record

anguss00:
@jonescc 

Fixes systest VIT. Thredds version 1.3.0 was expecting CRS to be included as a parameter, proxied requests were throwing 400s (!) as a result


anguss00:
Thanks @jonescc, will redeploy geoserver

PR #248, opened 2017-07-14
Always read/write variable data in bite size chunks to reduce memory …

jonescc:
…usage

Perhaps get a PO to confirm that chunked copies are actually working (i.e. no data is missing) before deploying to prod.


jonescc:
Required to subset bathymetry files without using an excessive amount of memory.

PR #249, opened 2017-07-19
Split layer parameter into workspace and layer when no workspace para…

akashisama:
…meter is received

Otherwise a null pointer is thrown later on for layers with a unique values regex, eg https://portal.aodn.org.au/search?uuid=60160b01-8ffc-45ce-a6f4-ee70ce391ec6


src/extension/layer-filters/src/main/java/au/org/emii/geoserver/extensions/filters/LayerFiltersService.java
```diff
@@ -148,6 +153,11 @@ public void uniqueValues(HttpServletRequest request, HttpServletResponse respons
         String layer = request.getParameter("layer");
         String propertyName = request.getParameter("propertyName");
 
+        if(workspace==null) {
```
lwgordonimos: Is it worth a slight change to avoid running that split method twice? e.g.
```java
        if(workspace==null) {
            String[] splitLayerArray = splitLayerName(layer)
            workspace = splitLayerArray[0];
            layer = splitLayerArray[1];
        }
```

src/extension/layer-filters/src/main/java/au/org/emii/geoserver/extensions/filters/LayerFiltersService.java
```diff
@@ -161,6 +171,19 @@ public void uniqueValues(HttpServletRequest request, HttpServletResponse respons
         }
     }
 
+    private String[] splitLayerName (String layerName) {
+        String[] splitArray = new String[2];
+
+        if(layerName.contains(":") ) {
+            splitArray[0] = layerName.split(":", 2)[0];
```
lwgordonimos: Since split returns an array, it shouldn't be necessary to manually assign elements? Just assign the whole array since it will always have length 2 here?

```java
splitArray = layerName.split(":", 2);
```

src/extension/layer-filters/src/main/java/au/org/emii/geoserver/extensions/filters/LayerFiltersService.java
akashisama: Yeah I was in two minds about that, feels slightly more readable with two calls I think, but also a bit wasteful. Have made change you suggest.

src/extension/layer-filters/src/main/java/au/org/emii/geoserver/extensions/filters/LayerFiltersService.java
akashisama: Yep, good catch, much better with one call.

PR #253, opened 2017-08-17
Provenance for WPS

pmbohm:
@pblain @jonescc The question remains; Is it necessary to include the Geonetwork metadata Url?
Geoserver currently has no reference to it? It could be passed in from Portal

@jonescc  - regarding Config
`params.put("gogoduckXml", config.getLayerConfigFilePath(layer));`
and 
`params.put("outputAggregationSettingsXml", settingsFile);`
are pretty bogus and need a look at!

@jones Why move the location of the templates? Are they intended to be part of the GeoserverDataDir?
Previously the templates were in 
`src/extension/wps/src/main/java/au/org/emii/wps/resources/provenance_templates....` and followed Freemarkers built in template lookup convention. I was following Tommy's `FilterConfigurationWriter` code which seems to involve less code.


craigj9:
re the metadata url:

I think its pretty desirable to be able to include it - provenance is about providing a historical record of the data and its origins.   @smancini?  

As to how, this is for MARVL - so the portal passing in the metadata Url won't help as they don't go via the portal.  Perhaps we need to look it up using catalogue-portal if its no longer maintained in geoserver-config?  I don't think we should be asking MARVL to pass in info we should be able to look up ourselves.

re config - no ones really clear on what makes sense here - its a suck it and see approach - if we are documenting the config used for the aggregation as an input then the files used and the date provides the possibility of tracking down exactly what the settings were at the time - @smancini?

re location of the templates - good question - I guess that's just where I had in mind that they should go similar to the layer ftl's - as I thought that changing the template should be easier than changing the code

jonescc:
re the metadata url:

I think its pretty desirable to be able to include it - provenance is about providing a historical record of the data and its origins.   @smancini?  

As to how, this is for MARVL - so the portal passing in the metadata Url won't help as they don't go via the portal.  Perhaps we need to look it up using catalogue-portal if its no longer maintained in geoserver-config?  I don't think we should be asking MARVL to pass in info we should be able to look up ourselves.

re config - no ones really clear on what makes sense here - its a suck it and see approach - if we are documenting the config used for the aggregation as an input then a link to the files used and the date provides the possibility of tracking down exactly what the settings were at the time.  We can look at how we can better specify the version when looking at versioning - @smancini?

re location of the templates - good question - I guess that's just where I had in mind that they should go similar to the layer ftl's - as I thought that changing the template should be easier than changing the code

smancini:
As Craig mentioned earlier, it is pretty important to include the Geonetwork metadata URL in the provenance record because it provides a way for the user to get more information about the process or the activity. Simon is using links to metadata in his example of provenance record for the MARVL software.

The config file and the information it contains is probably not going to be very useful to the end user, for now at least.
I think that for this task we are more interested in the mechanics of creating a provenance record and how we can link to the appropriate version of a file in GitHub.

akashisama:
@jonescc So your suggesting we'd use a search like this to find the point of truth from a layer name?
https://catalogue-portal.aodn.org.au/geonetwork/srv/eng/xml.search.summary?any=srs_ghrsst_l3s_1d_ngt_url%2Fsea_surface_temperature&sortBy=relevance&hitsPerPage=1&fast=index

jonescc:
Open to other suggestions -not completely happy with this - but portal config is where we keep this information.  

Searching using any works in this case and may be enough - if not we could add a specific search term to https://github.com/aodn/schema-plugins/blob/master/iso19139.mcp-2.0/index-fields.xsl 

If you remove the fast=index and add buildSummary=false you'll get a much more concise result.

jonescc:
Actually you need fast=index for the point of truth.

jonescc:
Ready for review

PR #261, opened 2017-10-31
Date formatting for CSV with header output

pmbohm:


ggalibert:
Is this in response to this issue: https://github.com/aodn/content/issues/297 ? 

IMOS standard string date format is: "yyyy-MM-ddTHH:mm:ssZ". All IMOS dates are stored in UTC. For example "2017-10-02T07:30:00Z".

pmbohm:
@ggalibert Thanks for that clarification. I was seeing varying values for the milliseconds so I assumed we would supply them.

jonescc:
Get an error running this locally:

31 Oct 13:57:10 ERROR [geoserver.ows] - 
java.lang.IllegalArgumentException: Illegal pattern component: T

jonescc:
You can testing this by creating a datastore that points to a 'production' featuretype (I chose soop_co2_trajectory_data) and use the csv-with-metadata WFS option on the layer preview page,

pmbohm:
@jonescc  Yes i'm reworking it now

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CSVWithMetadataHeaderOutputFormat.java
```diff
@@ -26,51 +14,168 @@
...
     }
 
     @Override
     protected void write(FeatureCollectionResponse featureCollection,
                          OutputStream output,
                          Operation getFeature) throws IOException, ServiceException {
         BufferedWriter w = new BufferedWriter(new OutputStreamWriter(output));
-        writeMetadata(featureCollection, getMetadataFeatureName(featureCollection), w);
-        this.csvOutputFormat.write(featureCollection, output, getFeature);
+        //writeMetadata(featureCollection, getMetadataFeatureName(featureCollection), w);
```
jonescc: We'll be needing this.

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CSVWithMetadataHeaderOutputFormat.java
```diff
@@ -26,51 +14,168 @@
...
 
     private Catalog catalog;
-    private CSVOutputFormat csvOutputFormat;
 
-    public CSVWithMetadataHeaderOutputFormat(GeoServer gs, Catalog catalog, CSVOutputFormat csvOutputFormat) {
+    public CSVWithMetadataHeaderOutputFormat(GeoServer gs, Catalog catalog) {
         super(gs, "csv-with-metadata-header");
         this.catalog = catalog;
-        this.csvOutputFormat = csvOutputFormat;
+        DateTimeZone.setDefault(DateTimeZone.UTC);
```
jonescc: This sets the default time zone for the JVM.  Best not to mess with this here - we are already specifying this as a system property when starting tomcat - specify the 'display' timezone when creating the datetime below - see comments below.

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CSVWithMetadataHeaderOutputFormat.java
```diff
@@ -26,51 +14,168 @@
...
+        }
+
+        w.write("\r\n");
+        NumberFormat coordFormatter = NumberFormat.getInstance(Locale.US);
+        coordFormatter.setMaximumFractionDigits(this.getInfo().getGeoServer().getSettings().getNumDecimals());
+        coordFormatter.setGroupingUsed(false);
+        FeatureIterator i = ((FeatureCollection)fc).features();
+
+        try {
+            for(; i.hasNext(); w.write("\r\n")) {
```
jonescc: Would 

    while (i.hasNext) {
        ...
        w.write("\r\n");
    }

be more readable?

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CSVWithMetadataHeaderOutputFormat.java
```diff
@@ -26,51 +14,168 @@
...
+    }
+
+    public String formatToString(Object att, NumberFormat coordFormatter) {
+        String value = null;
+
+        if (att instanceof Number) {
+            value = coordFormatter.format(att);
+        } else if (att instanceof Date) {
+            if (att instanceof java.sql.Date) {
+                value = utcDateString((Date)att);
```
jonescc: I'd leave this as serializeSqlDate.  By definition sql date is a date without time and without timezone so displaying it with a time and timezone doesn't sound right.  I don't think we'll ever use one of these - it sounds problematic.

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CSVWithMetadataHeaderOutputFormat.java
```diff
@@ -26,51 +14,168 @@
...
+        return value;
+    }
+
+    private String utcDateString(Date date) {
+
+        try {
+            String dateUTC = date.toString();
+            DateTime dt = IMOS_DATE_TIME_FORMATTER.parseDateTime(dateUTC);
+
+            DateTimeFormatter fmt = ISODateTimeFormat.dateTimeNoMillis();
```
jonescc: Not sure why we need to convert to a string and then parse the string.  Why not just create the datetime from the date as follows:

    DateTime dt = new DateTime( date, DateTimeZone.UTC );

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CSVWithMetadataHeaderOutputFormat.java
pmbohm: woops!!!!

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CSVWithMetadataHeaderOutputFormat.java
pmbohm: I tried `DateTime dt = new DateTime( date, DateTimeZone.UTC );` and it doesn't work.

This essentially is the problem i've spent all the time on, I would need help on this to make a change beyond what i have currently

pmbohm:
@jonescc Done (with Travis is doing its thing of course)

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CSVWithMetadataHeaderOutputFormat.java
```diff
@@ -70,7 +75,89 @@ protected void write(FeatureCollectionResponse featureCollection,
...
+                value = utcDateString((Date)att);
+            }
+        } else {
+            value = att.toString();
+        }
+        return value;
+    }
+
+    private String utcDateString(Date date) {
+        DateTime dt = new DateTime( date, DateTimeZone.UTC );
```
jonescc: Actually, probably need to do a null check here (returning an empty string).  Apparently, new dateTime(null) returns the current date/time.   See https://stackoverflow.com/questions/5042587/convert-from-java-util-date-to-jodatime

PR #265, opened 2017-12-06
Allow all protocols supported by the netcdf java library to be aggregated

jonescc:
by the core aggregator library.

* local filename (with a file: prefix or no prefix) for netCDF (version 3), hdf5 files etc
* OpenDAP dataset URL (with a dods:, dap4:, or http: prefix).
* NcML file or URL if the location ends with .xml or .ncml
* NetCDF file through an HTTP server (http: prefix)
* thredds dataset (thredds: prefix)

Build a standalone executable that can be used to aggregate files using the aggregator (I've been using this to test http aggregations)


gsatimos:
@jonescc I grabbed this from the board to PR. Is there an associated backlog or issue to do with this?

PR #266, opened 2017-12-15
Report size of files as they are added to the aggregation - could be …

ccmoloney:
…useful for stats gathering in future.


jonescc:
Branch is out  of date with base branch/check not successful.

jonescc:
Otherwise looks good.

PR #267, opened 2018-01-25
Prototype for z (depth) parameter subsetting

anguss00:
Linked PR https://github.com/aodn/aws-wps/pull/224
Implements depth (z dimension) as a ucar Range of monotonically increasing integers (based on the skeleton in aws-wps for handling z dimenion.

CARS datasets represent depth as a short variable, which is converted into a Range here (comments  and feedback on this conversion please).

Requires passing `DEPTH,0.0,100.0` as an addition to the subset; can be non-existant, in which case a null z-dimension is added to the subset.


anguss00:
This can be reviewed and merged now

PR #269, opened 2018-04-18
Capture current behaviour in integration tests before reorganising

jonescc:
Also, fix issue with Julian's custom code for loading data into the database for integration tests  (WTF!) - empty copy statement's caused statements to be skipped or loading to hang depending on where they occurred.


jonescc:
Superceded by https://github.com/aodn/geoserver-build/pull/270

PR #270, opened 2018-04-19
Read and write records in batches to restrict memory usage

jonescc:
Capture current behaviour in integration tests

Replace use of deprecated NetcdfFileWriteable with NetcdFilefWriter

Read and write records in fetchSize batches.


jonescc:
PS.  Now happily runs in 256m

PR #271, opened 2018-05-03
Only apply depth aggregation overrides where a z-dim exists

anguss00:
If z-dim exists, and a subset is not provided, use the max and min z-dimension of the dataset

### DIFF
--- a/src/aggregator/src/main/java/au/org/emii/aggregator/template/TemplateDataset.java
+++ b/src/aggregator/src/main/java/au/org/emii/aggregator/template/TemplateDataset.java
@@ -4,6 +4,7 @@ import au.org.emii.aggregator.dataset.AbstractNetcdfDataset;
 import au.org.emii.aggregator.dataset.NetcdfDatasetIF;
 import au.org.emii.aggregator.overrides.GlobalAttributeOverrides;
 import au.org.emii.aggregator.variable.NetcdfVariable;
+import au.org.emii.aggregator.variable.Bounds;
 import au.org.emii.aggregator.datatype.NumericTypes;
 import au.org.emii.aggregator.overrides.GlobalAttributeOverride;
 import au.org.emii.aggregator.overrides.AggregationOverrides;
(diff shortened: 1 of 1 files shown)


jonescc:
Nice!

anguss00:
Thanks @jonescc!

src/aggregator/src/main/java/au/org/emii/aggregator/template/TemplateDataset.java
```diff
@@ -158,9 +159,15 @@ public TemplateDataset(NetcdfDatasetIF dataset, AggregationOverrides aggregation
...
-        if (verticalRange != null) {
-            result.put("Z_MIN", Integer.toString(verticalRange.first()));
-            result.put("Z_MAX", Integer.toString(verticalRange.last()));
+        if (dataset.hasVerticalAxis()) {
+            if (verticalRange != null) {
+                result.put("Z_MIN", Integer.toString(verticalRange.first()));
+                result.put("Z_MAX", Integer.toString(verticalRange.last()));
+            } else {
+                Bounds verticalBounds = dataset.getVerticalAxis().getBounds();
+                result.put("Z_MIN", Integer.toString((int)verticalBounds.getMin()));
```
jonescc: Actually just noticed the (int)  -  looks to me like getBounds returns the min/max values as a double whereas verticalRange.first and verticalRangeLast returns the min/max index - probably should be returning the same type of thing - and I would have thought the value would be more relevant than the index for display in the global attributes.

jonescc: Tested this in prod and the values are coming through OK, but we shouldn't be coercing the value to an integer - this is CARS specific.

anguss00: Ah sorry @jonescc, missed your first comment. Doesn't verticalRange.first() and last() return the min/max value rather than the index? When I tested this in my own stack, that's what I got back in the attributes. But that doesn't seem to make sense... given that Range is a range of indexes. I'm a bit confused!

jonescc: First comment was incorrect are getting values.  However, there is an issue with converting to int. 
 For CARS values are int, other datasets may not be.   Conversion should happen when applying global attribute overrides if necessary.  This is an existing bug so I'm raising an issue for it.  No need to address here.

Issue #272, opened 2018-05-04
Vertical min/max assumed to be integer


jonescc:
Refer comments on https://github.com/aodn/geoserver-build/pull/271/files

We are currently assuming vertical axis values are integer when processing vertical subsets which works for CARS but will not work for datasets where this isn't the case.

PR #274, opened 2018-05-09
Issue 272

pmbohm:


src/aggregator/src/main/java/au/org/emii/aggregator/template/TemplateDataset.java
```diff
@@ -161,8 +161,8 @@ public TemplateDataset(NetcdfDatasetIF dataset, AggregationOverrides aggregation
 
         if (dataset.hasVerticalAxis()) {
             if (verticalRange != null) {
-                result.put("Z_MIN", Integer.toString(verticalRange.first()));
-                result.put("Z_MAX", Integer.toString(verticalRange.last()));
+                result.put("Z_MIN", verticalRange.start);
+                result.put("Z_MAX", verticalRange.end);
             } else {
                 Bounds verticalBounds = dataset.getVerticalAxis().getBounds();
                 result.put("Z_MIN", Integer.toString((int)verticalBounds.getMin()));
```
jonescc: Still using ints here!

src/aggregator/src/main/java/au/org/emii/util/ParameterRange.java
```diff
@@ -0,0 +1,14 @@
+package au.org.emii.util;
+
+
+public class ParameterRange {
+
+    public static String start;
+    public static String end;
+
+    public ParameterRange(String start, String end) {
+        this.start = start;
```
jonescc: Yep - like this, but why not use a double?

jonescc: So looking at the code this was being used internally only.  The aggregation worker was converting strings to data ranges, bounding boxes etc - thats probably where the vertical subset strings should be converted to a numeric range to be passed to the NetCDF Aggregator as per the other subset parameters.

src/aggregator/src/main/java/au/org/emii/aggregator/template/TemplateDataset.java
```diff
@@ -161,12 +161,12 @@ public TemplateDataset(NetcdfDatasetIF dataset, AggregationOverrides aggregation
...
-                result.put("Z_MIN", Integer.toString(verticalRange.first()));
-                result.put("Z_MAX", Integer.toString(verticalRange.last()));
+                result.put("Z_MIN", verticalRange.start.toString());
+                result.put("Z_MAX", verticalRange.end.toString());
             } else {
                 Bounds verticalBounds = dataset.getVerticalAxis().getBounds();
-                result.put("Z_MIN", Integer.toString((int)verticalBounds.getMin()));
-                result.put("Z_MAX", Integer.toString((int)verticalBounds.getMax()));
+                result.put("Z_MIN", Double.toString((double)verticalBounds.getMin()));
+                result.put("Z_MAX", Double.toString((double)verticalBounds.getMax()));
```
jonescc: Don't need to cast these to double - they are already.

src/aggregator/src/main/java/au/org/emii/util/DoubleRange.java
```diff
@@ -0,0 +1,13 @@
+package au.org.emii.util;
+
+public class DoubleRange {
+
+    public static Double start;
+    public static Double end;
```
jonescc: Did you mean public final?  static means these start and end fields will be shared between all instances of DoubleRange  e.g.

```
DoubleRange tenTo15 = new DoubleRange("10", "15");
DoubleRange twentyTo25 = new DoubleRange("20", "25");

System.out.println(tenTo15.start);
```

will print 20.0

final means they can never be changed after being set in the constructor and will be stored per instance if no static is used.

src/extension/wps/src/test/java/au/org/emii/wps/gogoduck/parameter/SubsetParametersTest.java
```diff
@@ -36,10 +36,10 @@ public void testValidTime() {
 
     @Test
     public void testValidDepth() {
-        SubsetParameters sp = SubsetParameters.parse("TIME,2009-01-01T00:00:00.000Z,2009-12-25T23:04:00.000Z;LATITUDE,-33.433849,-32.150743;LONGITUDE,114.15197,115.741219;DEPTH,0.0,100.0");
+        SubsetParameters sp = SubsetParameters.parse("TIME,2009-01-01T00:00:00.000Z,2009-12-25T23:04:00.000Z;LATITUDE,-33.433849,-32.150743;LONGITUDE,114.15197,115.741219;DEPTH,43.567,100.0");
 
-        assertEquals(sp.getVerticalRange().first(), 0);
-        assertEquals(sp.getVerticalRange().last(), 100);
+        assertEquals((Object) sp.getVerticalRange().start, 43.567);
```
jonescc: Do you need to cast to Object here?  doesn't seem to be required above for getLonMin

jonescc: assertEquals(double, double) is deprecated - they want you to use the one with the delta

pmbohm:
@jonescc Never heard of it.
Java to me is like macheting through a code jungle, without enough energy left to even remember why i'm slashing through the jungle.

jonescc:
Looks good but still need to apply some changes to aws-wps aggregation-worker which actually calls this, to use DoubleRange now instead of range.

src/aggregator/src/test/java/au/org/emii/aggregator/NetcdfAggregatorTest.java
```diff
@@ -144,7 +143,7 @@ public void testNonArrayDepthVariable() throws IOException, AggregationException
         AggregationOverrides overrides = AggregationOverridesReader.load(
             resourcePath("au/org/emii/aggregator/non-array-depth.xml"));
 
-        try (NetcdfAggregator netcdfAggregator = new NetcdfAggregator(
+            try (NetcdfAggregator netcdfAggregator = new NetcdfAggregator(
```
jonescc: indent prob

pmbohm:
fixes https://github.com/aodn/geoserver-build/issues/272

pmbohm:
@jonescc Fixed indentations

PR #281, opened 2018-09-26
Addresses aodn/backlog#941

jonescc:
Don't override the core PaletteKvpParser bean definition as this requires
context files to be loaded in the right order for it to work and this is outside our control

Instead use a simple parser which just returns the provided string for ncwms PALETTE request parameters


jonescc:
Self review failed.  Need to make the constructor more generic

jonescc:
Ready for review.

jonescc:
More problems to fix

jonescc:
Requires https://github.com/aodn/aodn-portal/pull/2728

PR #284, opened 2018-11-12
Skip over null rows when creating metadata headers

akashisama:
Fixes https://github.com/aodn/issues/issues/321


src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CSVWithMetadataHeaderOutputFormat.java
```diff
@@ -175,8 +175,10 @@ private void writeMetadata(FeatureCollectionResponse featureCollection,
             rs = stmt.executeQuery();
 
             while (rs.next()) {
-                w.write(rs.getString(1));
-                w.newLine();
+                if (rs.getString(1) != null) {
```
lwgordonimos: Could store the result of getString to avoid the duplicate getString method call?
e.g.

```java
            while (rs.next()) {
                String value = rs.getString(1);
                if (value != null) {
                    w.write(value);
                    w.newLine();
                }
}
```

akashisama: Yeah, a bit cleaner. Change made.

lwgordonimos:
I've built and tested in IntelliJ against prod data, and have confirmed that the file is not truncated when built from this branch.

PR #293, opened 2019-06-28
Use builder user instead of root

anguss00:


lwgordonimos:
HOME and JAVA_TOOL_OPTIONS are constants in the environment, so should be moved into the Dockerfile to keep the Jenkinsfile as lean as possible (see my commit on this branch). Same with other repos. 

I think I initially added them to Jenkinsfile for testing, but I think the Dockerfile is more correct going forward.

PR #308, opened 2020-05-27
Update jetty for local builds

pmbohm:
I needed this for local builds. I'm not sure how this could impact anything else, but it was required for me to be able to run it locally


pmbohm:
I was getting this error
https://stackoverflow.com/questions/45311295/error-scanning-entry-module-info-class-when-starting-jetty-server

pmbohm:
@jonescc bump

PR #309, opened 2021-01-19
Replace Travis CI with Github Actions

akashisama:
For https://github.com/aodn/issues/issues/940

It's failing in the same way as the travis build is: ```Error:  Non-resolvable import POM: Could not transfer artifact org.geoserver:geoserver:pom:2.15.0 from/to mvnrepository (http://mvnrepository.com/artifact/): Authorization failed for http://mvnrepository.com/artifact/org/geoserver/geoserver/2.15.0/geoserver-2.15.0.pom 403 Forbidden @ line 29, column 25```

Since this task is just to lift and shift travis builds to github actions I think we'll have to just accept it like that.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,8 @@ jobs:
     runs-on: ubuntu-16.04
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Cache Maven packages
         uses: actions/cache@v2
         with:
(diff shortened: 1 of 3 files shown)


lwgordonimos:
This looks like it could be the culprit? http://geotoolsnews.blogspot.com/2020/04/change-to-maven-repositories.html

lwgordonimos:
In addition to the geotools update, Central repository has also switched to HTTPS since this was last changed: https://blog.sonatype.com/central-repository-moving-to-https

lwgordonimos (approved):

PR #311, opened 2021-06-08
WSD Pivoting Mechanism

APaulic:
Backlog item: https://github.com/aodn/backlog/issues/3062

This is the implementation of the GeoServer part of the pivoting mechanism described [here](https://github.com/aodn/backlog/issues/3062#issuecomment-856267489).


APaulic:
@jonescc @ndhogan Would someone please be able to take a look at this PR for me?

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CsvPivotedFeatureCollectionSource.java
```diff
@@ -0,0 +1,185 @@
...
+
+        JSONParser parser = new JSONParser();
+        Set<String> foundColumnNames = new HashSet<>();
+        for (String sourceColumnName : pivotSourceColumnNames) {
+            try {
+                FeatureSource featureSource = ((TypeInfoCollectionWrapper.Simple) featureCollection)
+                        .getFeatureTypeInfo().getFeatureSource(null, null);
+                SimpleFeatureCollection features = (SimpleFeatureCollection) featureSource.getFeatures();
+
+                UniqueVisitor visitor = new UniqueVisitor(FF.property(sourceColumnName)) {
```
jonescc, line 90: We're after the unique keys - this will return unique json objects.   You'd need to use a postgres query to source the distinct key values with possible performance issues for large datasets - probably OK for WSD though.

jonescc: I guess this will work with the performance proviso - you'll most likely return all records.  But no real need to use a unique visitor.

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/DataDirectory.java
```diff
@@ -0,0 +1,93 @@
+/*
```
jonescc, line 1: Would be good if we could leverage the existing code instead of copying it.  Can just include filters as a dependency or separate out code to be reused into a separate module.

APaulic: Good point. I didn't know I could do that...

APaulic:
@jonescc I have now referenced the layer-filters classes instead of duplicating and query the db directly to get the column names.

APaulic:
@jonescc I just pushed an update to fetch the DB table name for the column names query.

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CsvPivotedFeatureCollectionSource.java
```diff
@@ -0,0 +1,195 @@
...
+            String attributeType = descriptor.getUserData().size() > 0 ? descriptor.getUserData().get("org.geotools.jdbc.nativeTypeName").toString() : "";
+
+            if (attributeType.equals(JSON_FIELD_TYPE)) {
+                pivotSourceColumnNames.add(attributeName);
+            } else if (!pivotConfig.getExcludedFields().contains(attributeName)) {
+                this.nonPivotColumnNames.add(attributeName);
+            }
+        }
+
+        // Read the first row to get JSON attr for column names
```
jonescc, line 83: No need to do this any more?

APaulic: Yes and no, the first row needs to be initialised but the column names do not need to be retrieved from it.

jonescc (approved):

PR #312, opened 2021-06-16
WSD - Pivot Prototype - User correct layer filters version

APaulic:
Backlog item: https://github.com/aodn/backlog/issues/3062

### DIFF
--- a/src/extension/csv-with-metadata-header/pom.xml
+++ b/src/extension/csv-with-metadata-header/pom.xml
@@ -56,7 +56,7 @@
       <dependency>
         <groupId>au.org.emii</groupId>
         <artifactId>layer-filters</artifactId>
-        <version>1.0.24</version>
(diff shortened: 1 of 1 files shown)


src/extension/csv-with-metadata-header/pom.xml
```diff
@@ -56,7 +56,7 @@
       <dependency>
         <groupId>au.org.emii</groupId>
         <artifactId>layer-filters</artifactId>
-        <version>1.0.24</version>
+        <version>1.0.25</version>
```
jonescc, line 59: Its currently 26 in master

APaulic: Haha, I'll get there eventually...

APaulic: This should be up to date now

PR #313, opened 2021-06-23
Update README to list all extensions included/customisations made

jonescc:


nspool (commented):
Is src/main/src/jetty/geoserver_data/gwc/metadata.properties intentionally left blank?

jonescc:
> Is src/main/src/jetty/geoserver_data/gwc/metadata.properties intentionally left blank?

Yes.  I just checked in changes geoserver made to the data directory used when starting it up using jetty.  GeoServer does this when there's missing config/config files - and generally only happens on a version upgrade or when installing a new plugin.  Better t include it in the repository so geoserver doesn't have to fix it up.

PR #317, opened 2021-10-06
Use correct default when json col is null

APaulic:
Backlog item: https://github.com/aodn/backlog/issues/3426


gsatimos (approved):

src/extension/csv-with-metadata-header/src/main/java/au/org/emii/geoserver/wfs/response/CsvPivotedFeatureCollectionSource.java
```diff
@@ -176,7 +176,7 @@ public boolean hasNext() {
                 }
             } else {
                 for (String key : this.pivotColumnNames.stream().sorted().collect(Collectors.toList())) {
-                    result.add("");
+                    result.add(this.pivotConfig.getDefaultValue());
```
mhidas, line 179: @APaulic @gsatimos Sorry, I think this was actually correct before. See updated requirements here: https://github.com/aodn/backlog/issues/3426#issuecomment-914740598
(apologies for the confusion, I wasn't able to update the backlog item directly...)

PR #319, opened 2022-01-19
wfs csv format dates

craigrose:
This will not include a timezone in the CSV if it is a java.sql.Timestamp


nspool (approved):

mhidas:
Is there a geoserver-build artefact that already includes this so I can test it in my stack?

